Issue #439, opened 2011-09-21
rsync_project symlink option


rizumu:
rsync has an option:
       -l, --links
              When symlinks are encountered, recreate the symlink on the destination.
 
It could be useful to optionally pass ``links=True`` to enable recreation of symlinks when using ``fabric.contrib.project.rsync_project``, otherwise symlinks will by default be ignored by rsync.

bitprophet:
rsync_project grew an arbitrary flags option, so I think this is no longer necessary.

PR #440, opened 2011-09-22
Following the example of rsync --delete, this should do the trick for rsync --links

rizumu:
Modified `~fabric.contrib.project.rsync_project` to honor the `--links` setting. re #439


bitprophet:
Think this is obviated by the `default_opts` kwarg that was added some time ago.

PR #370, opened 2011-12-20
put() empty remote_path arg ignores use of with cd():

GaretJax:
### Description

When using put() in this way:

```
with cd("/etc/apache2"):
    put('%s/httpd-vhosts.conf' % confs_dir, '', use_sudo=True)
    put('%s/apache2.conf' % confs_dir, '', use_sudo=True)
    sudo('mkdir logs')`
```

the empty string in put's second arg is supposed to resolve to the current working directory on the remote end. However, in this case it resolves to the user's home directory.

I believe the lines where this is determined are 397-400 in operations.py:

```
# Expand tildes (assumption: default remote cwd is user $HOME)
home = ftp.normalize('.')

# Empty remote path implies cwd
remote_path = remote_path or home
```

---

Originally submitted by **Matt DeBoard** (mdeboard) on **2011-06-29** at **03:32pm EDT**


elgow:
Almost another year gone by and this bug is still not fixed. At very least change the docs to inform people that they need to use '.' as the target dir if they want put() to work with cd().

PR #544, opened 2012-02-01
SFtp.put(): Preserve access and modification times

vitek:
Preserve file mtime and atime times useful when copying static to more than one frontend so browser can cache it depending on Last-Modified header.

Now utime() is called unconditionally for all local files (local_is_path is set).


mjeveritt:
See also updated version at #1705

PR #567, opened 2012-02-24
Bug when scaping quotes in sed

mviera:
Hi,

It seems there is a bug in sed when trying to escape quotes. I tried to replace 'HOST': '' to 'HOST': 'ip-address' and sed throws the following error:

[192.168.11.142] sudo: sudo -S -p 'sudo password:'  sed -i.bak -r -e 's/\'HOST\': \'\'/\'HOST\': \'192.168.11.142\'/g' /home/user/settings.py

Fatal error: sudo() received nonzero return code 1 while executing!

Requested: sed -i.bak -r -e 's/\'HOST\': \'\'/\'HOST\': \'192.168.11.142\'/g' /home/user/settings.py
Executed: sudo -S -p 'sudo password:'  sed -i.bak -r -e 's/\'HOST\': \'\'/\'HOST\': \'192.168.11.142\'/g' /home/user/settings.py

None

================== Standard output ======================================

bash: -c: line 0: unexpected EOF while looking for matching `''
bash: -c: line 1: syntax error: unexpected end of file
# 

It seems the best way to escape quotes in bash is the following: '\''

I already have sed working fine so I'm looking forward to find this fix accepted.

Cheers,
Manuel Viera.

### DIFF
--- a/docs/api/core/operations.rst
+++ b/docs/api/core/operations.rst
@@ -9,6 +9,6 @@ Operations
     .. autofunction:: get(remote_path, local_path=None)
     .. autofunction:: open_shell(command=None)
     .. autofunction:: put(local_path, remote_path, use_sudo=False, mirror_local_mode=False, mode=None)
-    .. autofunction:: reboot(wait)
+    .. autofunction:: reboot(wait=120)
     .. autofunction:: run(command, shell=True, pty=True, combine_stderr=True)
     .. autofunction:: sudo(command, shell=True, pty=True, combine_stderr=True, user=None)
--- a/docs/api/core/operations.rst
+++ b/docs/api/core/operations.rst
@@ -4,7 +4,6 @@ Operations
 
 .. automodule:: fabric.operations
     :members:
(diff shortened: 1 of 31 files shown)


bittecnologia:
Hi, today, with Fabric 1.10.2, I have exacly this problem.

I need to do exactly what @stuaxo did

       sed(filename='/etc/postgresql/9.3/main/postgresql.conf',
                              before="#listen_addresses = 'localhost' ",
                              after="listen_addresses = '192.168.1.2' ", 
                              use_sudo=True)

 The error
       out: bash: -c: line 1: syntax error: unexpected end of file

      Requested: sed -i.bak -r -e 's/#listen_addresses = \'localhost\' /listen_addresses = \'local\'/g'      "$(echo /etc/postgresql/9.3/main/postgresql.conf)"
     Executed: sudo -S -p 'sudo password:'  sed -i.bak -r -e 's/#listen_addresses = \'localhost\'  /listen_addresses = \'local\'/g' "$(echo /etc/postgresql/9.3/main/postgresql.conf)"


How can I make it works?

PR #802, opened 2012-12-20
Single quote handling in sed

shon:
So single quotes are used for wrapping the expressions.
So if expression uses signle quote(s) that needs to be escaped.
In case of single quote special escape is needed i.e. ' => '\''

TODO: add a test case
TODO: test can be as simple as sed(<path>, "'", '"')
TODO: Need to think about the fact that this is more shell specific than sed


grapemix:
+1.
In 2015, fanric 1.10.1, the single quote in sed problem is still there.

ps: I have not tested the patch

PR #824, opened 2013-01-16
Add a 'role' attribute to env.host_string

flupke:
This allows to know the 'current' role in a task.

Here is an example use case:

``` python
from fabric.api import env, sudo
from fabric.decorators import roles

env.dedupe_hosts = False

env.roledefs = {
    'web': ['webhost.example.com'],
    'worker': ['webhost.example.com', 'workers.example.com'],
}

def prod():
    env.prog_name = {
        'web': 'webapp',
        'worker': 'webapp-worker',
    }

def preprod():
    env.prog_name = {
        'web': 'webapp-preprod',
        'worker': 'webapp-worker-preprod',
    }

@roles('web', 'worker')
def restart():
    prog_name = env.prog_name[env.host_string.role]
    sudo('supervisorctl restart %s' % prog_name)
```

If env.host_string had no 'role' attribute we would have to guess it by searching env.host_string in env.roledefs. It would not work in this case because 'webhost.example.com' is present in both 'web' and 'worker' roles.

In the example above, the restart task will be executed 3 times, with env.host_string.role taking values ['web', 'worker', 'worker'].

I included a test in the commit, let me know if you accept the pull request and I will write documentation too.

### DIFF
--- a/fabric/task_utils.py
+++ b/fabric/task_utils.py
@@ -1,4 +1,4 @@
-from fabric.utils import abort, indent
+from fabric.utils import abort, indent, HostString
 from fabric import state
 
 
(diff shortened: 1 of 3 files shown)


mungayree:
flupke, Seem your example of prepod, prod apps do not work ad give attribute errors on latest fabric modules 1.11.1.

Issue #908, opened 2013-06-03
Support loading SSH keys from memory


calebgroom:
Currently [paramiko supports passing in SSH keys from memory](https://github.com/paramiko/paramiko/blob/master/paramiko/client.py#L234]) in addition to loading them from disk. Fabric should allow the user to set `env.key` and pass this value straight to paramiko.

Pull request coming.

PR #909, opened 2013-06-03
Support loading SSH keys from memory

calebgroom:
Issue #908.


YukSeungChan:
I saw this pull-request when i find way that use fabric with string ssh key.
Why this pull-request closed?

PR #927, opened 2013-07-08
Support "ProxyCommand none" ssh config

methane:


bitprophet:
Believe this got fixed at the Paramiko level a while back.

PR #932, opened 2013-07-11
contrib.files.upload_template should pass along temp_dir to put

nburlett:
In e7e96491, a temp_dir kwarg was added to fabric.operations.put. Since
fabric.contrib.files.upload_template makes use of fabric.operations.put,
it makes sense to add a temp_dir kwarg to upload_template and pass it
along to put.

### DIFF
--- a/fabric/contrib/files.py
+++ b/fabric/contrib/files.py
@@ -65,7 +65,7 @@ def first(*args, **kwargs):
 
 def upload_template(filename, destination, context=None, use_jinja=False,
     template_dir=None, use_sudo=False, backup=True, mirror_local_mode=False,
-    mode=None):
+    mode=None, temp_dir=''):
     """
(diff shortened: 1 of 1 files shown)


Andor:
Please, apply this merge request!

warsamebashir:
+1 Please this hurts

elgow:
This patch would offer at least a partial work-around to the design problem with put() when the sudo_user is not root. I hope you will merge it soon.  

The design for put() with sudo is fundamentally unsound for the general case. The problem is that the file gets transferred by sftp as the logged in user, but it's moved into its final place by the sudo_user. Unless the both the logged in user and the sudo_user have proper permissions on the temp directory and the file, the operation will fail at some point.  Probably what's necessary is to perform all remote operations as root and then chown the uploaded file to the sudo_user as the last step.

Issue #78, opened 2011-08-19
Add Tunneling Context to Fab


bitprophet:
### Description

This would be very helpful for me because it would make it very easy
to for example connect to a remote MySQL server using my command line
MySQL client.

We could use a "with" context statement, something like:

```
with tunnel(local=3307, remote=3306):
    local('mysql --port=3007 --host=localhost' mydb < db/dbdump.sql')
```

This would eliminate the need to upload mysql dump file to the server
just to be able to run import.

Another application could be for administering Cherokee web servers.

Cherokee web admin by default is only accessible from the server that
it's running on. So you want to access the admin you have to tunnel
into the server and access the admin interface using a local port.
This could also be simplified with this functionality.

```
with tunnel(local=9090, remote=9090):
   sudo('cherokee-admin')
   prompt('Stop cherokee admin?')
```

That last line would keep the tunnel open until it's closed by providing input.

---

Originally submitted by **Taras Mankovski** (tarasm) on **2009-11-02** at **09:30am EST**
### Relations
- Related to #38: Implement tunnelling

haridsv:
I recently needed this functionality to sync to a remote rsync server whose port is not directly reachable and found that the paramiko's forward.py demo code has sample code that I could use, so I came up with a solution that worked well for me and I submitted it as a patch for forward.py here: https://github.com/paramiko/paramiko/pull/504/

We could add `ForwardServer` from that patch and have a `local_tunnel()` that simply returns an instance of `ForwardServer`. As per recommendation from @bitprophet on the pull request, I will work on a patch for Fabric.

haridsv:
I actually didn't realize, but there is already a patch for `local_tunnel()`, though I am not fully sure about its status. What should I do?

bitprophet:
@haridsv If you're able, testing it out & mentioning that you successfully used it, on that patch (#939), would help, otherwise just have patience :) thanks!

naphthalene:
Is there any plan to resolve this issue? There are two outstanding pull requests that together resolve this, but they haven't been reviewed? Any way we could speed this along?
#939, #1218

bitprophet:
#939 is still in a release bucket, I just punted it from the upcoming 1.11 because I needed to cut *some* stuff, but it'll get priority for the next feature cycle.

PR #939, opened 2013-07-19
Add local_tunnel() context manager to forward a local port to the remote server (like 'ssh -L')

lfasnacht:
This pull request implements local_tunnel() context manager. This would close issue #78 (the first half was pulled in PR #821).

It was written the same way as remote_tunnel().

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ install:
   - pip install -e .
   # Install third-party multiprocessing on 2.5
   - "if [[ $TRAVIS_PYTHON_VERSION == '2.5' ]]; then pip install multiprocessing; fi"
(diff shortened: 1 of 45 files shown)


javiermatos:
This new feature works fine. I am using it in production with no issue at all (monkey patch). I'm willing to have it merged as soon as possible.

kaaveland:
The `select.error` that was mentioned and is fixed in https://github.com/fabric/fabric/pull/1218 is necessary. I have also had intermittent failures when repeatedly tunneling through the same host while changing the `fabric.api.env.hosts` in between, but I think that's probably a different issue. We have a team of ~20 people using this daily though and it seems to work fine with #1218 (monkeypatch).

PR #948, opened 2013-08-03
Handle connection failures due to server load

akitada:
SSHException with "Error reading SSH protocol banner" is raised when
sshd is unresponsive but Fabric treats it as an auth failure. This patch
fixes this by treating it as NetworkError.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -25,6 +25,10 @@ would have also been included in the 1.2 line.
 Changelog
 =========
 
+* :bug:`948` Handle connection failures due to server load and try connecting
+  to hosts a number of times specified in :ref:`env.connection_attempts
+  <connection-attempts>`.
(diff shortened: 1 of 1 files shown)


amanda11:
Hi, Could I get an update on this bug. I am having this same problem on a Solaris box with Fabric 1.10.1. The issue I have is that the network.py code that is in 1.10.1 has:
if e.__class__ is ssh.SSHException \
                and msg == 'Error reading SSH protocol banner':
                if _tried_enough(tries):
                    raise NetworkError(msg, e)
                continue

This works fine on my Scientific Linux box, as I do get raised "Error reading SSH protocol banner". BUT on my Solaris box then instead I get:
"Error reading SSH protocol banner[Errno 131] Connection reset by peer"

So as its doing a match with ==, rather than in as this bug fix suggested, it does not recognise this as a protocol banner.

Issue #949, opened 2013-08-06
Setting multiple shell environment variables desn't work properly under Windows with local()


jorisvddonk:
On Windows, when using shell_env to set environment variables, not all environment variables get set properly when using local(). Specifically, only the first variable gets set (any others are completely ignored) and the value assigned to that variable will be wrong as well (something like `"bar" SET baz="qux"` where `bar` is expected).

Test case:
https://gist.github.com/jorisvddonk/6162658

Results on Windows 7:
https://gist.github.com/jorisvddonk/6162668

Results on Linux (Ubuntu 12.04):
https://gist.github.com/jorisvddonk/6162673

bitprophet:
Rolling into PR.

PR #951, opened 2013-08-08
Fix for #949, Add && between SETs for Windows local env

mattPeloquin:
Added && between SET statements for Windows local case
Used glue replacement pattern for &&, so 

Windows local:  SET XXX="xxx" && SET YYY="yyy" && ...command

Other:  export XXX="xxx" YYY="yyy" && ...command


rlam3:
How do I set environment variables with fabric?

PR #1040, opened 2013-12-15
Enhanced layout for normal format task listing.

georgepsarakis:
Nicer, tabular display for normal task listing format. 
Tasks are colored green and full docstring is provided wrapped at the terminal width (using textwrap). 
Changed tests/test_main.py to reflect the changes.


bitprophet:
I'd rather roll this sort of change into 2.x, specifically the Invoke framework which forms the tasky bits for Fab 2. That said, totally dropping a link to it over in that tracker, because I like good ideas.

PR #1065, opened 2014-02-10
tiny change, fix incorrect reference to SSH config variable "ClientAlive...

hjwp:
...Interval"

Confusingly, the SSH client's keepalive parameter is "ServerAliveInterval" (ie, how often to check the server is alive), whereas "ClientAliveInterval" is something you would set server-side.

https://unix.stackexchange.com/questions/3026/what-do-the-options-serveraliveinterval-and-clientaliveinterval-in-sshd-conf

http://unixhelp.ed.ac.uk/CGI/man-cgi?ssh_config+5

http://unixhelp.ed.ac.uk/CGI/man-cgi?sshd_config+5

### DIFF
--- a/docs/usage/env.rst
+++ b/docs/usage/env.rst
@@ -393,7 +393,7 @@ The global host list used when composing per-task host lists.
 **Default:** ``0`` (i.e. no keepalive)
 
 An integer specifying an SSH keepalive interval to use; basically maps to the
-SSH config option ``ClientAliveInterval``. Useful if you find connections are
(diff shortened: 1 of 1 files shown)


bitprophet:
Wants backport to 1.10.x or thereabouts (I can do this next bugfix pass), otherwise looks good - thanks!

bitprophet:
Unclear why I only replied to this in 2017 when original was 2014...presumably somebody linked me to it or something 😊 

Been so long the original patch predates the site split; manually applied, will credit.

PR #1081, opened 2014-02-27
Add simulate initial login flag to sudo()

pcfens:
Allow `sudo -i` to be run with the `sudo` method by setting the `initial_login` flag.

This flag is useful when the .profile or other resource files need to be read by the shell before executing commands.

I attempted to write tests, but given this is the first time I've written tests for Python, I'm not sure that they work as intended.


foxx:
Any idea when this is getting merged? Just got stung pretty hard by this :/

PR #1099, opened 2014-03-14
- Added space char for better copy+paste 

guettli:
"Disconnecting from user@host .... ".

Before simple click on the user@host contained the three dots.


guettli:
.

PR #1121, opened 2014-04-20
Added `use_sudo` and `temp_dir` params to `operations.get()`. Fixes #700

### LINKED ISSUE #700: Feature: get() with sudo
It seems api.get() doesn't accept the use_sudo parameter like most other operations. It probably should. Not all files are world-readable.

xarg:
Also added some tests for the use_sudo + temp_dir combination for operations.put()

### DIFF
--- a/fabric/operations.py
+++ b/fabric/operations.py
@@ -411,7 +411,7 @@ def put(local_path=None, remote_path=None, use_sudo=False,
 
 
 @needs_host
-def get(remote_path, local_path=None):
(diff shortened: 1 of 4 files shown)


j-san:
Does not work yet for directories :)

example: `get('/data/media/', use_sudo=True)`

The problem comes from [sftp.glob](https://github.com/fabric/fabric/blob/master/fabric/sftp.py#L59)
The glob function does not suppert `use_sudo` param. Called from [fabric.operations.get](https://github.com/fabric/fabric/blob/master/fabric/operations.py#L567)

PR #1124, opened 2014-04-30
Fix many syntax errors on Python3.

alex:
No actual semantic fixes, so it still doesn't work, but maybe a first step.


asmeurer:
Why do you want to support Python 2.5? The last release was in 2011, and it's not even supported for security fixes any more. Supporting 2.5 and 3 in a single codebase is extremely annoying. I think there are a lot more people using Python 3 (or wanting to but can't because of fabric) than 2.5.

PR #1161, opened 2014-07-17
Added use_sudo as a kwarg to the reboot function.

btv:
This PR is to add a kwarg to reboot to allow ```run``` instead of just ```sudo``` within the reboot function.

No test written as no test currently exists for reboot.

Output from pylint is:
```
pylint2 -E operations.py 
No config file found, using default configuration
************* Module fabric.operations
E:827,54: Instance of 'list' has no 'endswith' member (but some types could not be inferred) (maybe-no-member)
E:828,53: Instance of 'list' has no 'endswith' member (but some types could not be inferred) (maybe-no-member)
```
Which isn't related to my change (though after looking at the code I can see why pylint is confused).

### DIFF
--- a/fabric/operations.py
+++ b/fabric/operations.py
@@ -1193,7 +1193,7 @@ def local(command, capture=False, shell=None):
 
 
 @needs_host
-def reboot(wait=120, command='reboot'):
+def reboot(wait=120, command='reboot', use_sudo=True):
     """
     Reboot the remote system.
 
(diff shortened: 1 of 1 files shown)


btv:
@bitprophet 
Sorry, I totally spaced on this commit. For the "Needs changelog/docs" flag, is there any documentation for adding to the docs?

bitprophet:
I really should split that label up, in this case it just meant "there's no changelog entry yet". Changelog is in `sites/www/changelog.rst` if you want to take a stab, otherwise I will add an entry when I merge. Gonna throw this into next feature release since this is pretty small. Thanks!

PR #1183, opened 2014-08-20
FIxing decorated tasks help

david-caro:
When using a decorated function (even using the wrapped decorator) it would
complain that it's not a python function when trying to get the arguments for
it. Now it will try also with the '**call**' method.


kaaveland:
I had hoped that this would also resolve https://github.com/fabric/fabric/issues/1406 but that is unfortunately not the case. For the fabfile in that example, this patch makes no difference. Maybe something else changed since then? I'm not sure which issue this is trying to solve, is it something completely different from #1406 ?

david-caro:
ping?

Issue #1203, opened 2014-10-02
should allow case insensitive match in fabric.contrib.files.contains()


xoul:
Hi,
maybe fabric.contrib.files.contains() within fabric/contrib/files.py should allow case insensitive match.

Since contains() uses egrep, where should be a way to pass it a "-i". 
Maybe a case_insensitive defaulting to false can be added:

fabric.contrib.files.contains(filename, text, exact=False, use_sudo=False, escape=True, shell=False, case_insensitive=False)

bitprophet:
Rolling into #1240

PR #1240, opened 2014-12-21
Feature #1203: Case-insensitive matching for contrib.files.contains()

cmattoon:
Adds keyword argument `case_sensitive=True` to contrib.files.contains().


bitprophet:
Thanks, looks good!

PR #1241, opened 2014-12-21
Fixes #1135. Command prefix is applied before shell_env.

### LINKED ISSUE #1135: Setting Shell_Env before running prefix
I would like to run and set my shell_env before the execution of the prefix. It looks like my shell env is never set in the example below:

````python

with shell_env(ENV1="TEST"):
    with prefix(". myapp"):
          run("echo $ENV2")
````

I expected my shell_env to be set but my prefix runs first. I know this because the prefix commands uses the environment variables and will prompt for them if not set?

cmattoon:
Reverses the order in which `shell_env` and `prefix` are applied to a command.

### DIFF
--- a/fabric/operations.py
+++ b/fabric/operations.py
@@ -904,7 +904,7 @@ def _run_command(command, shell=True, pty=True, combine_stderr=True,
 
         # Handle context manager modifications, and shell wrapping
         wrapped_command = _shell_wrap(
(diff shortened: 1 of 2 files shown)


bitprophet:
I think this will go into 1.11 since it's a "big" change for a bugfix.

Issue #1242, opened 2014-12-21
Impossible to do rsync_project() without passing "--rsh" option to rsync


cailloumajor:
In some cases (e.g. using a rsync daemon), it is necessary to not pass the "--rsh" option to the rsync command.
Solution proposed in PR #1243

bitprophet:
Rolling into the PR, thanks!

PR #1243, opened 2014-12-21
Fix for issue #1242

cailloumajor:
Allow to not pass "--rsh" option to rsync when using rsync_project()


bitprophet:
Merged into the 1.14 branch, next bugfix out soon.

PR #1246, opened 2014-12-30
missing a trailing slash in E.g. of docstring

tekapo:
The description says "ends with a trailing slash" but the example does **not** end with a trailing slash.


bitprophet:
Thanks!

PR #1250, opened 2015-01-08
Remove obsolete use of operator.isMappingType

rschiang:
Though still won't work in Python 3, this might be a little step forward as it's obsolete since 2.7.

### DIFF
--- a/fabric/main.py
+++ b/fabric/main.py
@@ -9,7 +9,7 @@ The other callables defined in this module are internal only. Anything useful
 to individuals leveraging Fabric as a library, should be kept elsewhere.
 """
 import getpass
-from operator import isMappingType
+from collections import Mapping
 from optparse import OptionParser
 import os
 import sys
@@ -361,7 +361,7 @@ def _sift_tasks(mapping):
     for name, value in mapping.iteritems():
         if _is_task(name, value):
             tasks.append(name)
-        elif isMappingType(value):
+        elif isinstance(value, Mapping):
             collections.append(name)
     tasks = sorted(tasks)
     collections = sorted(collections)


bitprophet:
We still need to support Python 2.6 though! If this code will work on 2.6 that's fine, but if not then we'll need to pass for now.

mathiasertl:
The documentation states that the `collections` module was added in Python 2.4, the Mapping type is not indicated to have been added later. It took me quite a while to get a python 2.6 shell (since there hasn't been a Debian/Ubuntu release supporting Python2.6 in almost five years), but here is what a 2.6 shell says:

```
root@pallene:/# python 
Python 2.6.5 (r265:79063, Apr 16 2010, 13:57:41) 
[GCC 4.4.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from operator import isMappingType
>>> from collections import Mapping
>>> isMappingType({}) == isinstance({}, Mapping)
True
>>> 
```

So you can merge this easily ;-)

naphthalene:
Will version 2.0 phase out Python 2.6?

bitprophet:
@naphthalene No, a nontrivial number of users still use it (frustratingly! but it's somewhere between 10-25% going by pip statistics) so 2.0 will support 2.6-3.x (90% of the codebase is actually http://pyinvoke.org which is 2.6+-3.3+ compat).

Now that I look at this issue though (I was about to anyways, 1.11 is imminent) I'm wondering why I said 2.6, because technically we still support **2.5** (according to our setup.py & docs). Sadly this isn't 2.5 compatible. While very few fresh installs are using 2.5 anymore, I'm loathe to cause backwards incompatible breakage for whomever's still stuck with it. So I think I'll actually close this :(

Issue #1200, opened 2014-09-27
Separate traceback display from debug output level


bitprophet:
tl;dr users shouldn't have to deal with debug output just to see tracebacks.

Adding tracebacks to the debug level was a convenient backwards compatible way to implement tracebacks partway through the dev cycle, but it conflates two concerns which are not always tied together.

We could add this in 1.x by adding a new output level just called eg `exceptions`, and either way it should be kept in mind for 2.x's logging setup.

stnatic:
Hi @bitprophet,

I could help by adding "exceptions" log level to the 1.x (master).

I looked for output.debug in the code and there are only a few occurrences. It seems like this (https://github.com/fabric/fabric/blob/a1e86d5435c07b8f83737d7523bcf74265227696/fabric/utils.py#L329) is the place where debug level is used together with printing an exception. Do you think that this is the only place (outside of tests and documentation) that needs to be adjusted?

PR #1258, opened 2015-01-20
Separate traceback display from debug output level

stnatic:
References issue #1200 

This pull request doesn't update the documentation. Could you let me know in the comments if this solution is OK and then I would adjust documentation?

### DIFF
--- a/fabric/utils.py
+++ b/fabric/utils.py
@@ -326,7 +326,7 @@ def error(message, func=None, exception=None, stdout=None, stderr=None):
     if func is None:
         func = fabric.state.env.warn_only and warn or abort
     # If exception printing is on, append a traceback to the message
-    if fabric.state.output.exceptions:
+    if fabric.state.output.exceptions or fabric.state.output.debug:
         exception_message = format_exc()
         if exception_message:
             message += "\n\n" + exception_message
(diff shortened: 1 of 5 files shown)


fabric/utils.py
```diff
@@ -325,9 +325,11 @@ def error(message, func=None, exception=None, stdout=None, stderr=None):
     import fabric.state
     if func is None:
         func = fabric.state.env.warn_only and warn or abort
-    # If debug printing is on, append a traceback to the message
-    if fabric.state.output.debug:
-        message += "\n\n" + format_exc()
+    # If exception printing is on, append a traceback to the message
+    if fabric.state.output.exceptions:
```
bitprophet: Let's preserve backwards compatibility by making this line say `if fabric.state.output.exceptions or fabric.state.output.debug` :)

bitprophet:
This seems reasonable to me, so yea, please update the docs & make a changelog entry. Thanks!

stnatic:
@bitprophet, thanks for feedback! I've added requested changes

bitprophet:
Sweet deal, thanks!

bitprophet:
I just updated the docs to adhere more cleanly with existing debug info, afb293f009395775390f4a811f5571fe2dcf1cad - FYI.

PR #1260, opened 2015-01-25
Hide blank lines when displaying stdout/err

Liudvikas:
This would avoid printing empty lines with prefix.


bitprophet:
Torn on this, there are times when users might expect empty lines (e.g. consider if one does `run('cat /a/file/containing/blank/lines')` or whatnot) but much of the time they *are* just noise.

I think I'd be more comfortable if this was implemented as an opt-in feature, e.g. testing `env.hide_blank_lines` which would default to `False`.

Issue #195, opened 2011-08-19
Paramiko doesn't support Kerberos


bitprophet:
### Description

Two people so far have asked in IRC about Kerberos support, which apparently is baked into the average SSH client by way of the GSSAPI protocol. User has Kerberos ticket locally somehow, which SSH then knows to forward to the remote end for authentication, in lieu of password or pubkey auth.

Unforunately Paramiko doesn't seem to have Kerberos support at this time (going by [this ServerFault question](http://serverfault.com/questions/103007/paramiko-and-gssapi), [this Paramiko Launchpad ticket](https://bugs.launchpad.net/paramiko/+bug/507391) and my trolling of the API docs). Which means Fabric doesn't either, save for the usual OOB trickery (if that even works with Kerberos; e.g. `local('ssh --something-kerberos %s' % env.host_string)` at the top of one's Fab tasks).

Users who need Kerberos should probably add their voice to the abovelinked Launchpad ticket.

I'm closing this as I create it, since there is no real action that can be taken on our end save for more hacking of Paramiko -- I just want it documented somewhere. Perhaps if my attempts on #72 go well I will become emboldened and try my hand at this too.

---

Originally submitted by **Jeff Forcier** ([bitprophet](https://github.com/bitprophet)) on **2010-07-15** at **09:58am EDT**
### Relations
- Related to #72: SSH key forwarding

---

Closed as _Wontfix_ on **2010-07-15** at **10:01am EDT**

PR #1261, opened 2015-01-28
Initial support for using Kerberos.

funkaoshi:
Support 3 additional parameters when connecting to hosts: gss-auth,
gss-deleg, gss-kex. These correspond to the equivalent in paramiko. It
does all the hard work.

### DIFF
--- a/fabric/network.py
+++ b/fabric/network.py
@@ -140,7 +140,7 @@ class HostConnectionCache(dict):
         Force a new connection to ``key`` host string.
         """
         from fabric.state import env
-        
+
         user, host, port = normalize(key)
         key = normalize_to_string(key)
         seek_gateway = True
@@ -449,7 +449,11 @@ def connect(user, host, port, cache, seek_gateway=True):
                 timeout=env.timeout,
                 allow_agent=not env.no_agent,
                 look_for_keys=not env.no_keys,
-                sock=sock
+                sock=sock,
+                gss_auth=env.gss_auth,
+                gss_deleg_creds=env.gss_deleg,
+                gss_kex=env.gss_kex,
+                gss_host=host,
             )
             connected = True
 
--- a/fabric/state.py
+++ b/fabric/state.py
@@ -141,6 +141,24 @@ env_options = [
         help="gateway host to connect through"
     ),
 
+    make_option('--gss-auth',
+        action='store_true',
+        default=False,
+        help="Use GSS-API authentication"
+    ),
+
+    make_option('--gss-deleg',
+        action='store_true',
+        default=False,
+        help="Delegate GSS-API client credentials or not"
+    ),
+
+    make_option('--gss-kex',
+        action='store_true',
+        default=False,
+        help="Perform GSS-API Key Exchange and user authentication"
+    ),
+
     make_option('--hide',
         metavar='LEVELS',
         help="comma-separated list of output levels to hide"


funkaoshi:
Oh, sorry, I didn't mean to start this right away. (I thought I was pulling against our forked repo.)

Anyway, does this look like the right approach? What sort of tests would you guys want here, if any. (I am just passing the various kerberos parameters from the env into paramiko.)

bitprophet:
Testing kerberos is actually super difficult as far as I can tell (we were unable to add meaningful tests to Paramiko's support) so I think we'd have to settle for just "are we passing the right parameters to the Paramiko level" perhaps.

Re: approach, it looks like an okay direction for now, this is basically what I was expecting (see https://github.com/fabric/fabric/issues/195#issuecomment-67093113). Thanks! Please comment again when you're ready for actual review (also note that I have a huge backlog D:)

kyrias:
@funkaoshi So you would consider this ready for merger/review now? If so I think I’ll give it a try.

funkaoshi:
@kyrias no one at the office has complained! It's a *very* small change.

funkaoshi:
(Right now for simplicity you can't pick a different host/port.)

bitprophet:
Eager to hear feedback from @kyrias but given how simple it is I'm gonna guess it works fine & any issues will be at the Paramiko level. Added to next feature release milestone.

funkaoshi:
Should I rebase this on top of master? (Or one of the release branches?) I'll add gss_host as an option, which will default to host if it's not set.

kyrias:
Yep, seems to work just fine for me.

bitprophet:
@funkaoshi Feature releases come off of master, so no need to move it to a release branch. Rebase to latest master if you want but the diff is pretty small & github isn't identifying any merge conflicts, so I doubt there's a problem at this point.

funkaoshi:
Sorry, one more thing. To use Kerberos tickets you need to install python-gssapi. Should this become a dependency of Fabric? I don't feel like it should, because how many people actually use Kerberos. On the other hand using `--gss-api` always fail if python-gssapi isn't available.

funkaoshi:
Also i'll be at PyCon too. What!

bennettaur:
This PR changed my life! A few years back ssh-keygen and I had an altercation which left me hospitalized for a month. Ever since then I've been unable to use keypair authentication and assumed my life was over as it was (password auth? HAH, here let me just package my identity up for you to use at will). But now with GSSAPI support I can fab deploy a new life without worry or fear ever again

But in all seriousness +1 on this
@funkaoshi I don't think python-gssapi should be a dependancy for fabric as it's an optional dependancy for paramiko for doer GSS-API stuff. Might be worth it to just mention it in the docs though

kyrias:
Explicitly erroring out/printing a warning when trying to use do GSS-API stuff without python-gssapi might be good, though the paramiko error, while less nice maybe, is sufficient I guess.

bitprophet:
TBH we should just make the paramiko error better, it's historically been kinda bluh at useful errors and it is one area I want to tidy up as I now own it :weary:

funkaoshi:
@bitprophet Agreed. Better to do things at the source, usually.

Are you guys (still) at PyCon?

funkaoshi:
Will someone else write the docs? I'm not sure where to start with that.

dahlke:
@funkaoshi I'm willing to write 'em up, don't think I can modify your PR though. Seems like the docs update should be bundled with the merge.

kyrias:
@dahlke You can send a PR to the sdelements/fabric basic-kerberos branch though.

funkaoshi:
Oh man it's been months. What do I need to do? Or how can I help you @dahlke? You could do a PR on top of my PR, yeah.

bitprophet:
Barf, yea it's been forever. We even semi-fixed-up GSSAPI after it broke again, in Paramiko. There's still some annoying confusion because of the issue documented in paramiko/paramiko/#584 but AFAIK if one has `python-gssapi` (vs `gssapi`) installed, it's functional right now.

If this basic "pass the vars through" approach still works with Paramiko 1.15/1.16 + `python-gssapi` I say we merge it and move from there. I can add a quick note to the install docs pointing at Paramiko's own note about GSSAPI install reqs, that will hopefully catch most users capable of reading.

bitprophet:
To be clear, what this means is I'd like @funkaoshi or somebody to:

* Update the PR to latest Fabric master (probably no large differences though...)
* Update local test environment to have at least Paramiko 1.15.4 if not the just-released 1.16.0
* Make sure your GSSAPI connection functions
* Let me know things look good

Then I will merge & tweak our install docs & it'll be ready for next Fab feature release (or I will just make one because big releases are dumb and a very bad habit).

funkaoshi:
Thanks. Let me rebase on top of master and see what's the what.

funkaoshi:
Will test things locally here and report back.

funkaoshi:
And based on my (non-exhaustive) testing it looks like this continues to work (assuming you have python-gssapi installed as well). We use this fork internally for some DevOps stuff, so I can ask them to update and see if anything breaks for them.

buckett:
It would be really nice to have this in a release, is there anything I can do to help get this merged?

bitprophet:
@buckett see https://github.com/fabric/fabric/pull/1261#issuecomment-154241099 - tl;dr additional "I tested this with a recent Paramiko release and a live Kerberos install and it worked for me" would be helpful.

That said, @funkaoshi gave a thumbs up so I mostly just need to find time to sprint through the 1.11 milestone that this is part of!

bitprophet:
Noticed one big problem with this, which is that it blows up on any Paramiko<1.15 (which is when Kerberos support landed). I'm fixing it up so it only submits those kwargs when they're set to non-`None` values (and updating the CLI/env defaults to be `None` instead of `False`).

This should mean enabling them works as before, but if one doesn't touch them, they're not submitted, and Paramiko e.g. 1.13 won't explode.

funkaoshi:
Ah, thanks for making the change.

PR #1264, opened 2015-01-30
Add check for environment variable to suppress colorized output

bergbrains:
Jeff,

This is the pull request that we discussed.

Eric

### DIFF
--- a/fabric/colors.py
+++ b/fabric/colors.py
@@ -22,13 +22,21 @@ Because these functions simply return modified strings, you can nest them::
 If ``bold`` is set to ``True``, the ANSI flag for bolding will be flipped on
 for that particular invocation, which usually shows up as a bold or brighter
 version of the original color on most terminals.
+
+It is now possible to disable color printing completely by setting the environment
+variable FABRIC_DISABLE_COLORS=True.
 """
 
+import os
 
 def _wrap_with(code):
 
     def inner(text, bold=False):
         c = code
+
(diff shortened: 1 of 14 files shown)


fabric/colors.py
```diff
@@ -22,13 +22,26 @@
...
 
     def inner(text, bold=False):
         c = code
+
+        try:
+            os.environ.get("FABRIC_DISABLE_COLORS")
+        except NameError:
+            os.environ["FABRIC_DISABLE_COLORS"] = None
+
+        if os.environ.get("FABRIC_DISABLE_COLORS") is not None:
```
bitprophet: Pro tip, you can actually condense all of this to simply:

``` python
if os.environ.get("FABRIC_DISABLE_COLORS", None):
    return text
```

This is because `os.environ` is a regular dict and `dict.get()` takes a 2nd param for a default value to apply if the key isn't in the dictionary. This is usually preferred over doing a try/except :)

bitprophet:
Cool, thanks @bergbrains!

I have some feedback, but I know you're still new to the process so I can do some of this on my end, simply let me know.

First, see the inline comments that will appear in here as I mark up specific parts of the diff.

Then, please add a changelog entry, see `sites/www/changelog.rst` and follow the format you see there, adding a new line item to the top.

bergbrains:
Good stuff, guys.  Thanks.  I'll make some changes and resubmit.

PR #1267, opened 2015-02-02
Update installing.rst

driscollis:
I noticed that this doc page said Fabric only supports Python 2.5 and 2.6 whereas Github states it also supports 2.7 (https://github.com/fabric/fabric)


driscollis:
Heck, even the main page (http://www.fabfile.org/) says it supports 2.7

bitprophet:
Thanks, good catch. (Also arguably a reason to just nix the mention from extra places, but meh.)

PR #1271, opened 2015-02-08
Changed env var check to a more elegant approach.  Thanks, @bitprophet!

bergbrains:
Jeff,

This is cleaned up and changed as per recommendations.


fabric/colors.py
```diff
@@ -22,13 +22,21 @@
...
 """
 
+import os
 
 def _wrap_with(code):
 
     def inner(text, bold=False):
         c = code
+
+        if os.environ.get("FABRIC_DISABLE_COLORS", None):
```
kaaveland: This should probably check if `is None`, if the intention truly is to have any value for `FABRIC_DISABLE_COLORS` work. It's possible to set it to the empty string and this would then disable the colors. So either the documentation or the code should be updated to stay in sync.

fabric/colors.py
```diff
@@ -22,13 +22,23 @@
...
 """
 
+import os
 
 def _wrap_with(code):
 
     def inner(text, bold=False):
         c = code
+
+        if os.environ.get('FABRIC_DISABLE_COLORS') is None:
```
kaaveland: I think maybe this contradicts the doc above now -- we'll only exit early here if `FABRIC_DISABLE_COLORS` is not set?

bitprophet: Jeez, yea. Thanks! I didn't even test this by hand :( Just did now to confirm. Pushed fix.

bergbrains: Great.  Thanks, Jeff.

I'm an old Perl hacker and was hoping to get a better take on how to
manipulate the fab run environment with command line or config file options
too.  haven't quite figured that out, but an env var is good for batch
processes.

Thanks again.

Eric

On Fri, Apr 8, 2016 at 11:35 PM, Jeff Forcier <notifications@github.com>
wrote:

> In fabric/colors.py
> <https://github.com/fabric/fabric/pull/1271#discussion_r59109647>:
>
> >
> >  def _wrap_with(code):
> >
> >      def inner(text, bold=False):
> >          c = code
> > +
> > +        if os.environ.get('FABRIC_DISABLE_COLORS') is None:
>
> Jeez, yea. Thanks! I didn't even test this by hand :( Just did now to
> confirm. Pushed fix.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/fabric/fabric/pull/1271/files/7cfdc3ce430add37a3206aa959b7ec55cd0e813e#r59109647>
>



-- 
eric berg
eberg@bergbrains.com

PR #1279, opened 2015-02-20
Swapped non-ASCII character for (c) to fix broken build

kelvinn:
The build is failing because of a non-ASCII character. The file's encoding could be updated, or the character could be swapped, which is included in this commit.


bitprophet:
Starting to think I need to re-enable emails for Travis, I totally missed this. Nice catch! Thank you.

Issue #1032, opened 2013-12-03
Using @hosts or @roles decorator meddle with fab -d taskname output


shaftoe:
It's a small bug, but still...

alex~$ cat fabfile.py
from fabric.api import *

@task
@roles('fake_role')
def broken_task(my_argument=True):
    '''Arguments is ok only if @roles is missing'''
    local('Hello world!')

alex~$ fab -d broken_task
Displaying detailed information for task 'broken_task':

    Arguments is ok only if @roles is missing
    Arguments:

alex~$ sed -i '/fake_role/d' fabfile.py

alex~$ fab -d broken_task
Displaying detailed information for task 'broken_task':

    Arguments is ok only if @roles is missing
    Arguments: my_argument=True

Issue #971, opened 2013-08-29
`fab -d task` shows incorrect arguments when using @parallel @serial decorators 


avishai-ish-shalom:
This is because the implementation of the decorator uses `@wraps` instead of `@decorator` which does not copy arguments metadata

bitprophet:
Rolling into #1286

PR #1286, opened 2015-03-04
WrappedCallableTask.__details__() support for multi-decorated objects

mvk:
- fixes fabric/fabric #971
 - fixes fabric/fabric #1032

### DIFF
--- a/fabric/tasks.py
+++ b/fabric/tasks.py
@@ -177,7 +177,10 @@ class WrappedCallableTask(Task):
         return getattr(self.wrapped, k)
 
     def __details__(self):
-        return get_task_details(self.wrapped)
+        orig = self
+        while 'wrapped' in orig.__dict__:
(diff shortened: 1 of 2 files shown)


mvk:
The failure happens after the tests I've added passed. this is false positive on tests failure :(

bitprophet:
Travis seems to have broken things across the board for our test suite, so no that's not your fault. Thanks for the submission :) Also, what's up with #1287, is that supposed to supersede this one or was it a mistaken push?

mvk:
Hi.
Not entirely, but the last commit out of these 3 is probably a Travis
phantom fix :)

bitprophet:
Confirmed a downstream dev dependency broke our tests, thankfully I also run taht project now so I fixed it :3 kickstarted travis again to see what happens.

mvk:
Glad to hear. Do you want me to do anything or you're OK with cherry
picking?

Issue #1222, opened 2014-11-11
Bad quoting on `run`


k-bx:
I'm not sure if it's related to fabric or underlying library, but I see a misbehavior in quoting a command. To check that everything (doesnt) work, I created few test files. Pay no attention to actual command that's being run (i.e. `cabal`). 

Important part is that my end-goal is to run `cabal configure --flags="-flag1 -flag2"`

Files:

1. fabfile.py

```python
from fabric.api import *


env.hosts = ['root@0.0.0.0:49175']
env.password = 'docker'

def test1():
    local("CONFIGUREARGS=\"--flags=\\\"-flag1 -flag2\\\"\" make build",
          shell='/bin/bash')


def test2():
    local('CONFIGUREARGS="--flags=\\"-flag1 -flag2\\"" make build',
          shell='/bin/bash')

def test3():
    run('CONFIGUREARGS="--flags=\\"-flag1 -flag2\\"" make build',
        shell='/bin/bash')
```

2. Makefile
```
build:
        cabal configure $(CONFIGUREARGS)

.PHONY: build
```

3. Run a docker container with some image with ssh-server (in my case it's [this ghc image](https://github.com/k-bx/docker-ghc-7.8)):

```
sudo docker run -d -it -P my/ghc
```

This command runs a docker container with ssh-server working. List running containers with `sudo docker ps`, get it's name, get ssh-port with `sudo docker port <container_name> 22`. Put that port to fabfile instead of mine.

4. Put Makefile with same contents as in 2nd item here to `~/Makefile` in running container (not necessarily, since command will fail even before that)

If you're still with me, we are ready to reproduce the bug.

If you run `fab test1` or `fab test2`, you can see it happily runs desired command at the end. But! If you run `fab test3` -- it will crash with:

```
deploy@builder:~/tmp/fabricquote$ fab test3
[root@0.0.0.0:49175] Executing task 'test3'
[root@0.0.0.0:49175] run: CONFIGUREARGS="--flags=\"-flag1 -flag2\"" make build
[root@0.0.0.0:49175] out: -flag2\" make build: -c: line 0: unexpected EOF while looking for matching `"'
[root@0.0.0.0:49175] out: -flag2\" make build: -c: line 1: syntax error: unexpected end of file
[root@0.0.0.0:49175] out:


Fatal error: run() received nonzero return code 1 while executing!

Requested: CONFIGUREARGS="--flags=\"-flag1 -flag2\"" make build
[...]

PR #1291, opened 2015-03-09
Fixed shell escaping slashes (for nested quotes)

rendaw:
I believe this is the same issue as #1222

### DIFF
--- a/tests/test_operations.py
+++ b/tests/test_operations.py
@@ -1,24 +1,17 @@
 from __future__ import with_statement
 
-import os
-import shutil
-import sys
-import types
 from contextlib import nested
 from StringIO import StringIO
 
-import unittest
-import random
-import types
-
-from nose.tools import raises, eq_, ok_
-from fudge import with_patched_object
+from nose.tools import ok_
+from fudge import with_fakes, Fake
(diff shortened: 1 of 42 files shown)


tests/test_operations.py
```diff
@@ -303,6 +303,14 @@ def test_shell_escape_escapes_backticks():
     eq_(_shell_escape(cmd), "touch test.pid && kill \`cat test.pid\`")
 
 
+def test_shell_escape_nested_doublequotes():
+    """
+    _shell_escape() escapes nested double-quotes
+    """
+    cmd = 'cd "\\"Application\\" Support"'
+    eq_(_shell_escape(cmd), 'cd \\"\\\\\\\"Application\\\\\\\" Support\\"')
```
ploxiln: you can use `r'...'` quoted python strings to make this test more readable (though I'm not sure in what version of python they were introduced) e.g.

```
>>> print('cd \\"\\\\\\\"Application\\\\\\\" Support\\"') ; \
... print(r'cd \"\\\"Application\\\" Support\"')
cd \"\\\"Application\\\" Support\"
cd \"\\\"Application\\\" Support\"
```

PR #1297, opened 2015-03-15
Some Python 3 fixes

fmalina:
These popped up while Running setup.py install for fabric. There is more.


fmalina:
Oh, I am sorry. I understand you are going through a big rewrite.

PR #1302, opened 2015-04-08
Add --always-exclude-hosts option.

Quuxplusone:
This is exactly like --exclude-hosts, except that it is respected
by nested calls to `execute()`. So if you have a machine that is
definitely down, and you just want to skip any connection that would
normally be made to that machine, you can set `env.always_exclude_hosts`
at the very top of your supertask and know that it will be respected
all the way down through the tree of subtasks (unless of course they
explicitly reset `env.always_exclude_hosts` themselves).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,5 +32,4 @@ notifications:
       - "%{repository}@%{branch}: %{message} (%{build_url})"
     on_success: change
     on_failure: change
-    use_notice: true
   email: false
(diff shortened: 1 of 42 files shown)


sivy:
Looks like a good addition. Can we take care of this Sphinx error/warning in the travisci build?

> WARNING: Title underline too short.

>    \`\`always_exclude_hosts``
>    \-----------------

sivy:
Also, is there a good reason not to make this default behavior? Why introduce a new option rather than make `--exclude-hosts` simply apply to nested operations?

@bitprophet

bitprophet:
Yea I don't get why the existing setting *doesn't* apply to nested `execute()` driven things, IIRC those are supposed to preserve (most of) the 'parent' environment's settings, or merge them, depending.

So either it wasn't/isn't easy or possible due to the design of `--exclude-hosts` somehow (implemented too high up in the cli module?) or it was simply an oversight and needs adding to the code that bridges "parent" and "child" fabric environments.

Definitely agree with @sivy that we should investigate that first :)

Issue #1294, opened 2015-03-11
append() does not respect current content if line contains '<'


ecksun:
I noticed that if I run append() several times with the same content it also appends that line several times, instead of just one time. That is, append() is not idempotent.

```
If text is already found in filename, the append is not run, and None is returned immediately.
```

The following small test script can reproduce the issue:

``` python
from fabric.contrib.files import append

def run():
      append('testfile', 'some < pipe')
```

This issue seems to be similar to #341

The error has to do with the '<' symbol.

The problem seems to boil down to the regex that is sent to egrep in contains():
https://github.com/fabric/fabric/blob/master/fabric/contrib/files.py#L360

printing egrep_cmd gives me the following for the above testcase:

```
egrep "^some\ \<\ pipe$" "$(echo testfile)"
```

However the correct regex would be:

```
egrep "^some\ <\ pipe$" "$(echo testfile)"
```

The incorrect regex seems to be generated by the `_escape_for_regex` function.

AndreCimander:
+1, still encountering this bug in 1.12, our poor servers are restoring their iptables rules way too many times x)

ecksun:
I have provided the fix in the initial description, you can probably create a patch from that and submit it. I'm no longer using fabric so will likely not fix this myself :(

bitprophet:
Fixed in #1559!

PR #1304, opened 2015-04-13
Fix for #1294

cmattoon:
In favor of keeping it simple, I went with the approach of replacing pipes with a (hopefully) unique string. This allows `re.escape` to ignore the string completely. Then, just sub it back in at the end.

Since `<`, `>`, `&` and `|` can be used in both regex and as redirects/pipes, the function distinguishes the two by whether or not there's a leading and trailing space. (`|regex|` vs `cat file | more`).Things like ` 2>&1` would still be escaped, but that could be handled easily by hard-coding common redirects, or using regex. I'm not sure I'm thrilled with either. 

Apologies for the messy test; I'll try and get something more appropriate (and that reflects the underlying issue with `contains` and `append` more closely) later.

### DIFF
--- a/tests/test_contrib.py
+++ b/tests/test_contrib.py
@@ -4,7 +4,7 @@ from fabric.operations import local
 import os
 
 from fabric.api import hide, get, show
-from fabric.contrib.files import upload_template, contains
+from fabric.contrib.files import upload_template, contains, _escape_for_regex
 from fabric.context_managers import lcd
 
 from utils import FabricTest, eq_contents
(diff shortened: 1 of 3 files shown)


tests/test_contrib.py
```diff
@@ -129,3 +129,23 @@ def _upload_template_obeys_lcd(self, jinja, mirror):
...
+            r"cat file2 >> file3": r'cat\ file2 >> file3',
+            r"ls > dirlist 2>&1": r'ls > dirlist\ 2\>\&1',
+            r"<html><head></head><body>":r'\<html\>\<head\>\<\/head\>\<body\>',
+            r"Mailto: <cmattoon@cmattoon.com>": r'Mailto\:\ \<cmattoon\@cmattoon\.com\>',
+            r"cat file3 | more": r'cat\ file3 | more'
+            }
+
+        for (text, expected) in valid.iteritems():
+            assert expected == _escape_for_regex(text)
+
```
sivy: Should lose the extras whitespace.

Issue #701, opened 2012-08-06
Sudo() captures incorrect output when user initially enters wrong password


egmanoj:
Consider the following sudo call using Fabric:

```python
def ls():
    output = sudo('ls /tmp/', user = 'someone', shell = True)
    print "Output:", output
```

Execute this using fab and enter a *wrong* password the first time you are prompted by sudo. You'll see an error message and sudo will ask you for your password again. Enter the correct password the second time. The session looks something like this:

```shell
$ fab ls
[user@server] Executing task 'ls'
[user@server] sudo: ls /tmp/
[user@server] out: sudo password: 

[user@server] out: Sorry, try again.
[user@server] out: sudo password: 

[user@server] out: hsperfdata_someone

Output: Sorry, try again.

Done.
````

Note how the value of output is the first message printed by sudo when the wrong password is entered ("Sorry, try again."). It doesn't contain the subsequent correct output ("hsperfdata_someone").

PS: I asked this as a question in StackOverflow and didn't get any answers. Feel free to reject this if I'm doing something wrong.

PR #1311, opened 2015-04-13
Fixes the processing of try again prompts.

toastdriven:
In trying to hunt down #701, I encountered the bug that the chomping of the try again prompt was wrong. 
This fixes that.

**Before:**

```
['\r', '\n', 'S', 'o', 'r', 'r', 'y', ',', ' ', 't', 'r', 'y', ' ', 'a', 'g', 'a', 'i']
```

**After:**

```
['S', 'o', 'r', 'r', 'y', ',', ' ', 't', 'r', 'y', ' ', 'a', 'g', 'a', 'i', 'n', '.', '\r', '\n']
```


bitprophet:
Thanks, I gotta make sure I reproduce & test this on my end, and shit out a changelog entry.

bitprophet:
Hrm, can't recreate this on my end, if you remember what you did @toastdriven can you drop me a hint on how to reproduce?

I also note that the change removes the setting of the `try_again` variable so I'm worried that might goof up other use cases. That might need another look.

PR #1319, opened 2015-04-18
Fix determination of whether or not the task decorator was invoked.

contracode:
Fixes `@task('foo')` decorator usage.


bitprophet:
This looks all right at first blush, but can you provide a concrete use case the released behavior is breaking (that this fixes)? I also feel like I've seen this before, so you might want to look for prior PRs.

contracode:
@bitprophet - At present, this does *not* fix any released behavior, since any custom Task object must specify the `task_class` keyword, by design. If that were to change, however, this extra checking would be necessary.
 
Searched prior PRs and didn't find anythng.

PR #1321, opened 2015-04-21
Replaced deprecated isMappingType

xbello:


bitprophet:
We support Python 2.5 and 2.6 in addition to 2.7, which is when `collections.Mapping` was added - so we can't use this. Sorry! Thanks anyway :)

PR #1323, opened 2015-04-24
Fix for #1294

cmattoon:
In favor of keeping it simple, I went with the approach of replacing pipes with a (hopefully) unique string. This allows re.escape to ignore the string completely. Then, just sub it back in at the end.

Since <, >, & and | can be used in both regex and as redirects/pipes, the function distinguishes the two by whether or not there's a leading and trailing space. (|regex| vs cat file | more).Things like 2>&1 would still be escaped, but that could be handled easily by hard-coding common redirects, or using regex. I'm not sure I'm thrilled with either.

Apologies for the messy test; I'll try and get something more appropriate (and that reflects the underlying issue with contains and append more closely) later.

(Removed whitespace & rebased on `master`)


bitprophet:
Subsumed into #1559.

PR #1326, opened 2015-04-28
Gateway support to contrib/project.py for rsync

drasch:


bitprophet:
Thanks! Adding to milestone.

drasch:
glad to hear it.  let me know if I can help.

drasch:
For future improvement, I think there's now a better way to do this with SSH without using netcat, but this method is very reliable and continues to work for me.

PR #1334, opened 2015-05-27
Added killing local subprocess after timeout seconds to fabric.api.local

kblw:


komlenic:
@bitprophet I ran into a scenario today that is solved precisely by this solution.  'run' accepts a timeout arg, and it would be helpful if 'local' did similarly.  Should/can this PR be accepted and included in the next 1.x release?

bitprophet:
This is small enough that yea, sure, it can live in a 1.x feature release.

vadviktor:
Is this feature hoing to be released anytime really soon? This would help me not having to have to bash together complex shell pipelines when some nasty process times out.

PR #1359, opened 2015-08-04
Expose the Changelog

jvanasco:
I couldn't find the changelog until I did quite a bit of searching, because it's hidden in a subdirectory.

I propose the following patch -- it simply adds a CHANGELOG.rst file on the top level (where one would expect it to me) that identifies the actual changelog, and gives a link to the HTML version on fabfile.org


bitprophet:
Thanks!

PR #1361, opened 2015-08-05
Fixes #1348: handle % chars in get() in a better way

### LINKED ISSUE #1348: SFTP transfer crashes on paths with "%" characters
When transferring files from a host with `get()`, an exception is triggered if any path has a "%" code in it.

    Fatal error: get() encountered an exception while downloading
    Underlying exception:
        %d format: a number is required, not dict

To reproduce, create a directory on the remote host with % in the file:

    ssh remote 'mkdir -p top/%a; touch top/%a/%d'
    mkdir out

The run the following script to try to retrieve this dir:

    from fabric.api import settings, get
    with settings:
       get('top/*', 'out')

The issue is in the % expansion within sftp, where it passes a dict of args before calling abspath() in sftp.py:

            local_path = os.path.abspath(escaped_path % path_vars )

bspink:
Improved the regex for escaping the % characters.

I put this against 1.5 as that's the earliest branch I could find this in (the contributions guide said to do this), and it still persists in master. The changelog file has moved location in master, so if you want me to change anything let me know. :)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -25,6 +25,9 @@ would have also been included in the 1.2 line.
 Changelog
 =========
 
+* :bug:`1348` Fix get() to not break when string format chars appear in files
+  on the remote server. Thanks `@natecode` for the bug report and `@bspink`
+  for the fix.
 * :release:`1.5.5 <2013-12-24>`
 * :bug:`956` Fix pty size detection when running inside Emacs. Thanks to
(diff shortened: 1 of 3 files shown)


bitprophet:
Had to cherry-pick due to changelog moving between 1.5 and 1.10, but this is merged to 1.10 and up now. (Release soon.)

Issue #1365, opened 2015-08-23
Exception subclasses should not be shown in "fab -l"


mattvonrocketstein:
To reproduce, create a fabfile.py with contents "class Foo(Exception): pass", then "fab -l" will list it as an available command.  

I suppose this happens because callable(Foo) is True, but since using an Exception as a fabric command is nonsense, I would expect this to generate an error like "Fatal error: Fabfile didn't contain any commands!" just like if I had a fabfile consisting only of contents "Foo=[]", 

For real fabfiles, this clutters up the namespace and it's annoying to have to do the "from module import SomeException as _SomeException" workaround.

bitprophet:
Gross. Thanks for the catch.

mattvonrocketstein:
No problem.  Since we agree it's a bug now, I'll volunteer to create a patch.

bitprophet:
Rolling into PR.

PR #1372, opened 2015-09-03
fix and test for issue 1365: exception subclasses should not be recog…

mattvonrocketstein:
…nized as fabric commands

### DIFF
--- a/fabric/main.py
+++ b/fabric/main.py
@@ -9,6 +9,7 @@ The other callables defined in this module are internal only. Anything useful
 to individuals leveraging Fabric as a library, should be kept elsewhere.
 """
 import getpass
+import inspect
 from operator import isMappingType
 from optparse import OptionParser
 import os
@@ -125,6 +126,7 @@ def is_classic_task(tup):
             callable(func)
             and (func not in _internals)
             and not name.startswith('_')
+            and not (inspect.isclass(func) and issubclass(func, Exception))
         )
     # Handle poorly behaved __eq__ implementations
     except (ValueError, TypeError):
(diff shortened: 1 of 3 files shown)


fabric/main.py
```diff
@@ -125,6 +126,7 @@ def is_classic_task(tup):
             callable(func)
             and (func not in _internals)
             and not name.startswith('_')
+            and not (inspect.isclass(func) and issubclass(func, Exception))
```
kaaveland: I'm wondering to myself if it makes sense that any class is a task? I'm wondering if maybe we should just refuse classes. Suppose I want to generate some csv from a bunch of servers and I put the following in my file: `from csv import CsvWriter, Dialect` -- with current master, fabric will say `    Dialect     Describe an Excel dialect.` in `fab -l`.

Maybe we shouldn't remove only exceptions here?

mattvonrocketstein: I think what you say is totally reasonable, and would be all for excluding all classes and not just exception subclasses.  FWIW, I made this change for exceptions only because it seemed like a good combination of useful/safe.  It's useful because, for quick fabfiles, I'm betting exceptions are imported more often than classes. It's safe because even though there's probably one maniac out there that has [some crazy work flow](https://xkcd.com/1172/) that depends on the existing fabfile beahviour with respect to classes, *surely* no one is counting on the existing Exception behaviour, right?

kaaveland: Fair enough, I do agree that some crazy person could be using `__init__` as a task and I agree this seems less likely to break for anyone.

bitprophet: Class-based tasks are actually a definite (if minority) use case, so the limitation of "it's an Exception" should stay, IMHO.

bitprophet:
Dropping #1365 here for linkage.

bitprophet:
Cherry-picked to 1.10 branch as 4e1fd8357d93a792cba492c920c5810aabc1d0c1 - closing & releasing soon, thanks!

PR #1387, opened 2015-10-08
Support for passing a regex to put and get operations and filter by matching names

haridsv:
Added match_pattern and match_excludes optional parameters to both operations for fine control on what names are copied.


haridsv:
@bitprophet Could you take a look at these changes?

gtback:
@haridsv: If I remember right, @bitprophet is going to be out of touch for a few weeks. You might not hear back for a while.

haridsv:
@gtback Is anyone else going to take responsibility in his absence?

gtback:
@haridsv: Not that I know of. I'm not sure who else, if anyone, has commit access to this repo.

bitprophet:
I'm torn on this, it's a moderately sized change and we're trying to avoid additional stability hits in the run-up to 2.0. Will leave open for now in case I change my mind. Thanks!

haridsv:
@bitprophet This change doesn't come into picture unless the new keyword arguments are used, so shouldn't impact the existing functionality (the proof is in the passing of all existing tests, untouched). If you think otherwise, I can look for ways to make it even less intrusive. Also, if you see any gap in unit tests please point out and I will add them.

PR #1388, opened 2015-10-12
allow trailing newline in Jinja2

aleiphoenix:
Jinja2 by default will trim a single trailing newline.

http://jinja.pocoo.org/docs/dev/templates/#whitespace-control

Most time we will need this trailing newline IMHO :)

### DIFF
--- a/fabric/contrib/files.py
+++ b/fabric/contrib/files.py
@@ -68,7 +68,7 @@ def first(*args, **kwargs):
 
 def upload_template(filename, destination, context=None, use_jinja=False,
     template_dir=None, use_sudo=False, backup=True, mirror_local_mode=False,
-    mode=None, pty=None):
+    mode=None, pty=None, keep_trailing_newline=False):
     """
     Render and upload a template text file to a remote host.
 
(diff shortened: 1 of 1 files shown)


bitprophet:
Can you clarify when that kwarg was added to Jinja? tl;dr if this is going to change behavior unexpectedly for folks, I'd rather punt on it for now :) thanks!

aleiphoenix:
Ah, how could I never this change will break someone else. I've changed the code, adding a 'keep_trailing_newline' kwargs, because it's `False` default to jinja2 env.

datashaman:
I too have this problem. Files in /etc/cron.d must end in a newline (for some stupid reason), otherwise they do not take effect. My templates have that explicit newline, which is stripped out on deployment. So basically, deploying disables all application crons. Not ideal. :)

bitprophet:
@datashaman Hah, gross. thanks for the useful feedback! (Not sarcasm.) Tossing this into a release bucket since it's small & decently sized impact for some use cases.

buzzijavier:
I just ran into this issue today, with cron.. merge timeframe??

datashaman:
If you're stuck with this issue ATM, it's easy enough to add another step
in your fabfile in the meantime:

    sudo('echo "" >> /etc/cron.d/myfile')

PR #1389, opened 2015-10-14
fix nested executions using non-standard ports

HarryWeppner:
Nested executions using non-default ports did not work as env.port was prioritized over hostname and ssh config derived port numbers.

See test_tasks.py:TestExecute.test_nested_execution_with_explicit_ports

### DIFF
--- a/fabric/network.py
+++ b/fabric/network.py
@@ -140,7 +140,7 @@ class HostConnectionCache(dict):
         Force a new connection to ``key`` host string.
         """
         from fabric.state import env
-        
+
         user, host, port = normalize(key)
         key = normalize_to_string(key)
         seek_gateway = True
@@ -294,10 +294,11 @@ def normalize(host_string, omit_port=False):
     # values)
     r = parse_host_string(host_string)
     host = r['host']
+
     # Env values (using defaults if somehow earlier defaults were replaced with
     # empty values)
     user = env.user or env.local_user
-    port = env.port or env.default_port
+
     # SSH config data
     conf = ssh_config(host_string)
     # Only use ssh_config values if the env value appears unmodified from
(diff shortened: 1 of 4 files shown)


HarryWeppner:
Hi there,

do you have an eta when you'll be able to review and merge this pull request?

Thanks & cheerio, Harry.

bitprophet:
I'm not comfortable making this type of change at this point in 1.x's lifecycle (see http://www.fabfile.org/roadmap.html) - the chance for breaking existing-known-to-work functionality feels a bit too high :( thanks, and sorry!

HarryWeppner:
Hi,

that's a bummer because I

* tried to keep code changes to an absolute minimum
* avoided larger refactoring to stay within the existing style
* made sure it passes all existing tests
* added a new test to avoid future regressions

The _only_ aspect this commit disambiguates is the priority of port information, which is reflected in a comment.

Working around this bug is painful and so is adding a custom patch for vanilla installations.

I'd appreciate it if you could reconsider and review the changes once more. On my end, I'll promise to find time to contribute to the 2.0 roadmap.

Thanks & cheerio, Harry.

bitprophet:
Thank you for taking the time for a reasoned & calm response, I appreciate it!

To expand my initial reply, I'm not knocking the quality of your work or the caution you took (again, it's appreciated!). However, the history of code (or at least, all projects I've maintained) is rife with well-meaning, well-tested, small changes which nonetheless broke previously unseen/untested use cases. The more critical the code path (`network.connect` is as critical as it gets), the more of those pitfalls come into play.

So every bugfix is a risk assessment balancing act of "how many people are encountering this bug?" (I haven't seen many  other reports that seem to map to this) vs "what's the chance of this bugfix causing other bugs?" (higher than I'd like).

Compound that with the previously-noted conservative approach that I'm applying to Fabric 1 as 2 looms, and you get my original "ehh this changes how `connect()` handles port numbers, that's scary" reaction.

----

All that said, maintainership is a constant battle between "this could break shit!" and "people can always just roll back!" and I'm trying to cut more, smaller releases - meaning the cost of rollback should be lower than it might otherwise. So I will think about this some more and make another judgement call soon. Thanks again.

bitprophet:
Upon reflection, let's do it after all. Thanks @HarryWeppner. Adding to release milestone.

HarryWeppner:
great - thanks for reconsidering! In the (unlikely) event it breaks something I'll be around to help...

HarryWeppner:
Sorry, accidentally published a commit for a unrelated pull request here, which has been separated to pull request #1398 .

fabric/network.py
```diff
@@ -305,17 +306,29 @@ def normalize(host_string, omit_port=False):
...
+
     if omit_port:
         return user, host
+
+    # determine port from ssh config if enabled
+    ssh_config_port = None
+    if env.use_ssh_config:
+        ssh_config_port = conf.get('port', None)
+
+    # port priorities (from highest to lowest)
```
kaaveland: This sort of thing seems to me as something that should be documented somewhere. Maybe something like this should go into the docstring here?

tests/test_tasks.py
```diff
@@ -417,6 +417,47 @@ class MyTask(Task):
...
         execute(mytask)
 
+    @server(port=2200)
+    @server(port=2201)
+    def test_nested_execution_with_explicit_ports(self):
+        """
+        nested executions should work with defined ports
+        """
+
+        def sub_task_one():
```
kaaveland: Matter of personal taste I guess, but I think the main task would read more clearly if this was named something like `assert_port_is_2201`

bitprophet:
Merged, including addressing Robin's concerns. Thanks both!

PR #1390, opened 2015-10-16
Make it explicit that the sample needs to be saved as fabfile.py

haridsv:
For folks who are new to Fabric (or coming back to after a long time), it can take sometime to figure out about the default file ``fabfile.py``, so just mentioning that explicitly.


bitprophet:
Thanks!

Issue #196, opened 2011-08-19
open_shell() doesn't do readline too well


bitprophet:
### Description

E.g. up/down history occasionally gets messed up, requiring an extra up arrow to get going; ctrl-a seems to operate across multiple lines sometimes; etc.

Do some rigorous testing to see just what isn't working and try to figure out why, in case it is something that can be fixed easily.

Not _terribly_ concerned since this is just a nice-to-have thing that's not usually required, given that run() can be interacted with...

---

Originally submitted by **Jeff Forcier** ([bitprophet](https://github.com/bitprophet)) on **2010-07-17** at **05:42pm EDT**
### Relations
- Related to #7: Improved prompt detection and passthrough

mbasaglia:
I've found this issue to affect `run` and `sudo` when they start any interactive command that uses readline. 

It seems like the first up key is not synced with the output and all subsequent keys are offset by that (ie: the key is sent correctly but the output is one key stroke behind)

PR #1398, opened 2015-11-13
fix escape sequences in open_shell

HarryWeppner:
In the context of open_shell with sys.stdin connected to a channel,
cursor movements appear "wonky", which can really mess up the user
experience.

An initial escape sequence is not captured correctly and only partially
sent across the channel resulting in no effect. A repeated key press
yields the correct sequence.

I suspect this is caused by a mismatch between select (mapped to a
syscall) and reading from the (buffered) sys.stdin. Dropping to the
lower-level os.read resolves this issue entirely.

NOTE: this may explain https://github.com/fabric/fabric/issues/196

### DIFF
--- a/fabric/io.py
+++ b/fabric/io.py
@@ -2,6 +2,7 @@ from __future__ import with_statement
 
 import sys
 import time
+import os
 import re
 import socket
 from select import select
@@ -232,7 +233,7 @@ def input_loop(chan, using_pty):
             have_char = (r and r[0] == sys.stdin)
         if have_char and chan.input_enabled:
             # Send all local stdin to remote end's stdin
-            byte = msvcrt.getch() if win32 else sys.stdin.read(1)
+            byte = msvcrt.getch() if win32 else os.read(sys.stdin.fileno(), 1)
             chan.sendall(byte)
             # Optionally echo locally, if needed.
             if not using_pty and env.echo_stdin:


HarryWeppner:
@bitprophet this problem may have been lingering for a while - this became a real problem when using fabric to run through an initial OS install where dialog prompts aren't very forgiving.

bitprophet:
Wondering if this is related to things I encountered in v2 (via Invoke, which redoes all of the local command running machinery), e.g. [this comment and subsequent ones](https://github.com/pyinvoke/invoke/issues/289#issuecomment-166156531).

Looking at the PR here, it'll affect not only `open_shell` but `run` as well, so I'm pretty loathe to change it up given Fab 1.x is intended to be stable these days :(

I'm going to close this for now but ensuring it's linked to that Invoke ticket, and will try & remember to revisit it when Fabric 2's alpha is moved to the public repo. (Ideally, this sort of thing - given it's re: local input reads - will behave the same regardless of whether the command is being run locally or remotely.)

HarryWeppner:
@bitprophet not sure I can parse your comment correctly. I'd be happy to explain _why_ the bug existed and what the fix does. Please let me know if you need further information.

We've been running with this fix in production ever since without any problems. `open_shell` is definitely broken without this fix.

Given this is a very isolated and clearly understood change, I'd appreciate for you to reconsider. I hope that `stable` doesn't mean that bugs will _not_ be fixed!?

Cheerio, Harry.

bitprophet:
What I meant is that the switch from `file_obj.read` to `os.read` will affect input reading for *all* command execution via `run`/`sudo`, not just use of `open_shell`, and I've run into enough wacky corner cases regarding this exact type of switch (`os.read` vs `file_obj.read`) in other codebases that I'm worried it'll cause mysterious issues.

As noted in that other ticket where we had this convo, it comes down to the balancing of risk, and risking issues in very commonly used stuff (`run`) in favor of less commonly used stuff (`open_shell`) feels off.

Further, the v2 codebase and its underpinnings (Invoke) have a lot of work in this specific area (it's why I'm aware of there being subtle differences between the different read methods) and I'd prefer to centralize those hijinx instead of spreading them across multiple major versions.

Issue #1346, opened 2015-06-25
trailing whitespace in ssh user leads to bad ssh connection


twomice:
This is similar to #305 and #619, which relate to the ssh host, whereas this relates to ssh user.

# Summary:
A trailing space in the ssh user will be passed on to the ssh connection process, leading to a failed connection attempt.  This is true for trailing spaces in the User value for a host in .ssh/config, leading to the confusing situation in which `ssh example.com` connects properly, but `fab mytask -H example.com` fails.

# To reproduce
ssh-agent is running with my key for the remote server, as you'll see in the results of the `ssh` command below.

```
$ cat fabfile
from fabric.api import *
env.use_ssh_config = True

@task
def test_fab():
    run('pwd')
$ cat ~/.ssh/config
Host mie
  Hostname A.B.C.D # redacted IP address
  # note the trailing whitespace on the next line:
  User allen 
$ ssh mie
Last login: Thu Jun 25 14:38:13 2015 from W.X.Y.Z
allen@mie:~$ exit
logout
Connection to A.B.C.D closed.
$ fab test_fab -H mie
[mie] Executing task 'test_fab'
[mie] run: pwd
[mie] Passphrase for private key: 
```

If I remove the trailing space from the User value in .ssh/config, there's no problem connecting with fabric:
```
$ cat ~/.ssh/config
Host mie
  Hostname A.B.C.D # redacted IP address
  # No more trailing whitespace on the next line:
  User allen
$ ssh mie
Last login: Thu Jun 25 14:38:13 2015 from W.X.Y.Z
allen@mie:~$ exit
logout
Connection to A.B.C.D closed.
$ fab test_fab -H mie
[mie] Executing task 'test_fab'
[mie] run: pwd
[mie] out: /home/allen
[mie] out: 


Done.
```

I also tried adding `env.user = "allen"` to fabfile.py, allowing for a successful connection. But changing that to `env.user = "allen "` results in the same type of failure shown above. 

Perhaps adding a trailing space inside the quotations when setting env.user is "asking for trouble" in that it should be fairly obvious and will also fail consistently.  Trailing spaces in .ssh/config, however, are not as obvious and will also not fail when connecting directly with `ssh`.

bitprophet:
Thanks for the report!

bitprophet:
This actually might need fixing in Paramiko but we'll figure that out when we get to this.

abednarik:
[...]

PR #1399, opened 2015-11-16
Strip whitespace in username

abednarik:
Related to #1346

### DIFF
--- a/fabric/state.py
+++ b/fabric/state.py
@@ -49,7 +49,7 @@ def _get_system_username():
             import win32security
             import win32profile
             username = win32api.GetUserName()
-    return username
+    return username.strip()
 
 def _rc_path():
     """


bitprophet:
Thanks!

bitprophet:
Poking this further, I think the "best" fix is actually to do the strip at connection time. The change here only affects default usernames derived from the OS, which is certainly *a* way a bad username could creep in (though I don't know if any major OSes allow such usernames!) but most of the bad input is likely to come from SSH config files (as per the report in #1346).

I'm about to merge my own tweak to accomplish that, but still giving you credit in the changelog. Thanks!

bitprophet:
Dug deeper at this and the core issue is actually SSH config parsing (I don't think systems actually allow trailing spaces in local usernames, which this PR here would fix - please comment if you're finding otherwise, @abednarik!) which was fixed in Paramiko as https://github.com/paramiko/paramiko/issues/499 - will be out in next Paramiko bugfix release cycle, then even users on older versions of Fabric will be able to get the fix by updating Paramiko.

PR #1428, opened 2016-02-04
Fabfile discovery: loading package with *.pyc only

### LINKED ISSUE #1427: Fabric can't load package with *.pyc only
Fabric's fabfile discovery algorithm only treats directories with an `__init__.py` file in them as legal Python packages.  It refuse to load any directories with an `__init__.pyc` file only (without `__init__.py`) in them , which are also legal Python packages.

oldsharp:
Fabfile discovery: loading package with *.pyc only

    As the doc (sites/docs/usage/fabfiles.rst, in "Fabfile discovery")
    says, Fabric is capable of loading Python modules (e.g. fabfile.py) or
    packages (e.g. a fabfile/ directory containing an __init__.py).
    Starting from version 0.9.2, Fabric has been capable of loading package
    fabfiles.

    However, Fabric's fabfile discovery algorithm only treats directories
    with an __init__.py file in them as legal Python packages.  It refuse
    to load any directories with an __init__.pyc file only (without
    __init__.py) in them, which are also legal Python packages.

    Fabric's current fabfile discovery algorithm works well for most cases.
    But in some cases, for example, suppose someone would like to release a
    tool, which consist of several Fabric tasks, in such a format that it
    only contains *.pyc files, i.e. he/she does not want to release the
    source code at all.  In such a case, users could run into trouble since
    Fabric refuse to load packages which only have an __init__.pyc in them.

    This commit gives Fabric the ability to load package fabfiles with
    __init__.pyc file only.

    This resolves #1427.

### DIFF
--- a/fabric/main.py
+++ b/fabric/main.py
@@ -75,9 +75,10 @@ def _is_package(path):
     """
     Is the given path a Python package?
     """
+    _exists = lambda s: os.path.exists(os.path.join(path, s))
     return (
         os.path.isdir(path)
-        and os.path.exists(os.path.join(path, '__init__.py'))
+        and (_exists('__init__.py') or _exists('__init__.pyc'))
     )
 
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -3,6 +3,7 @@ from __future__ import with_statement
 import copy
 from functools import partial
 from operator import isMappingType
+import os.path
 import sys
 
 from fudge import Fake, patched_context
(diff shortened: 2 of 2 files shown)


oldsharp:
Not sure this should be treated as a bug or a feature, so I branch it off `master`.

I will push changelog changes if necessary.

kaaveland:
As far as I can tell, the travis failure here is unrelated. Looks good :+1:

oldsharp:
@bitprophet Re-fire a build and now it's passed.  Please have a check at your convenience.

bitprophet:
Seems this got past me when it was first posted, labeled and added to bugfix milestone - looks good at a glance. Thanks!

bitprophet:
Confirmed test fails w/o impl and passes w/ impl 👍

PR #1443, opened 2016-03-12
For consistency, do not use a relative import

bildzeitung:
This line, when Fabric is used with a plugin architecture that re-writes Python's internal import mechanism, results in a stack trace.

Making this line consistent with the rest of the codebase solves this issue.

### DIFF
--- a/fabric/network.py
+++ b/fabric/network.py
@@ -396,7 +396,7 @@ def connect(user, host, port, cache, seek_gateway=True):
         Whether to try setting up a gateway socket for this connection. Used so
         the actual gateway connection can prevent recursion.
     """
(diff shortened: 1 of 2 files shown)


kaaveland:
As far as I can tell, the build failure is a sphinx issue, which wasn't touched here. 

I think @bitprophet has his own procedures wrt to bumping the version, so you should probably leave that out. You can add yourself to https://github.com/fabric/fabric/blob/master/sites/www/changelog.rst

bildzeitung:
My bad -- I didn't think those would be picked up in the pull. I'll revert those or resubmit this in a branch like a grownup. Apologies.

PR #1462, opened 2016-05-01
Handle Fabric-level assumptions about PyCrypto's use

alex:
This is in prep for bumping the paramiko dependency, where Cryptography is used instead of PyCrypto.

### DIFF
--- a/fabric/decorators.py
+++ b/fabric/decorators.py
@@ -6,7 +6,10 @@ from __future__ import with_statement
 import types
 from functools import wraps
 
-from Crypto import Random
+try:
+    from Crypto import Random
+except ImportError:
+    Random = None
 
 from fabric import tasks
 from .context_managers import settings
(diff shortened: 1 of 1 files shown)


bitprophet:
Can you rebase this against...let's see, the 1.10 branch, please? If we do go this route I'd like to put it out for all supported lines. thanks!

alex:
@bitprophet as in, close this PR and file a new one against that branch?

bitprophet:
Booo the base-branch changing stuff isn't terrifically clean :( just acts like a straight up merge or something so all of $NEWER_BASE_BRANCH commits end up applied to $OLDER_BASE_BRANCH.

Doesn't actually mattress for something this small but I was hoping that new GH feature would save me the manual cherry-picking/rebasing. Oh well. Will do that momentarily...

bitprophet:
Actually waffling on whether this should go into the older releases or not. Changing the Paramiko dep in bugfix releases feels like a no-no; should be trivial for users to upgrade to 1.14 if we do it there. Merging this without also changing the Paramiko dep feels pointless. So yea, meh. Let's just do it in 1.14. I do waste too much time on older release lines sometimes >_>

Issue #935, opened 2013-07-14
Remote Globbing is Limited


jklukas:
The docstring for fabric.operations.get doesn't disclose the limitations of the remote globbing logic.  It states:

> remote_path is the remote file or directory path to download, which may contain shell glob syntax, e.g. "/var/log/apache2/*.log"

The given example is valid, but the globbing is done only on the filename, not on the directory structure.  I propose this be changed to "may contain shell glob syntax _for the filename piece_, e.g.".

Or better yet, implement globbing on the directories too.  That would be awesome!

PR #1470, opened 2016-06-02
Avoid creating local files for "matchless" patterns

shashurup:
When there is no match for a pattern fabric creates empty local file
with pattern name.

### DIFF
--- a/fabric/operations.py
+++ b/fabric/operations.py
@@ -559,11 +559,11 @@ def get(remote_path, local_path=None, use_sudo=False, temp_dir=""):
 
         try:
             # Glob remote path if it's a directory; otherwise use as-is
-            if (
-                ftp.isdir(remote_path)
-                or '*' in remote_path or '?' in remote_path
-            ):
+            is_pattern = '*' in remote_path or '?' in remote_path
+            if ftp.isdir(remote_path) or is_pattern:
                 names = ftp.glob(remote_path)
+                if is_pattern and names == [remote_path]:  # No matches for our pattern
+                    names = []
             else:
                 names = [remote_path]
 


mmalchuk:
+1 this fix is really needed
also closes https://github.com/fabric/fabric/issues/935

bitprophet:
Actually, this is not quite the right fix, it hides the error that is naturally raised on no glob matches.

The real bug is in `fabric.sftp.SFTP.glob`.

Wrote a test to prove this issue exists & another one to enforce the "errors on glob miss" behavior; will merge shortly.

bitprophet:
Also I don't see that this does fix #935 really, it doesn't make directory globbing function or document that it's not implemented =/

bitprophet:
Put in a "better" fix that updates `sftp.glob` to return empty list when no matches, then an explicit test for that situation in `get` which raises a similar `IOError` to the old behavior.

I also tested OpenSSH's `sftp` to make sure that it does itself error out when globs fail to find anything - and they do. (Otherwise I might have just omitted the explicit `IOError`...)

bitprophet:
Grump, that breaks other stuff (like [grabbing whole directories](https://travis-ci.org/fabric/fabric/jobs/154303989#L705)). So glad I am getting rid of this horrible get/put shite in 2.0 :( Reimplementing SFTP/rsync has been nothing but pain.

It *looks like* I can address the other breakage by removing the `isdir` test entirely; this all smells like organically decomposing code. Certainly, grabbing paths which are directories, still works recursively now, with no need to use globbing to obtain their contents (which I suspect was the original goal of that statement).

shashurup:
I agree that returning just empty list from glob() would be resonable. However it is considered natural glob behaviour to return the pattern itself when there are no matches, take a look, for example, at bash.

Issue #1469, opened 2016-06-02
errors/aborts in parallel mode should stop all execution.


aravind:
Current behaviour: In parallel mode, fabric moves on to the next host in the list when a task fails. In serial mode, fabric does the right thing and stops when it gets an error/abort.

New behaviour (maybe toggled by a flag): In parallel mode, it would be nice if fabric could stop when it notices that one of the tasks failed on a host.

Example use case: Let's say I am deploying something to 20 servers, two at a time. If a task failed on one of the servers, chances are it will fail on all of them. I'd rather not risk pushing out a bad deploy to all 20 boxes and abort the whole thing (as early as possible) on noticing the first failure.

The user in this comment https://github.com/fabric/fabric/issues/457#issuecomment-218124012 seems to be referring to a similar situation.

goosemo:
To clarify, you'd like to have the option to halt on first error, when running in parallel mode?

aravind:
@goosemo yup, that would be ideal.. but if that were not possibble halt as soon as possible.

So in the 20 server situation above..

1. fabric kicks stuff on box 1 and box 2.
2. something fails on box 1. fabric aborts on box 1.
3. fabric continues with box 2 since it's already processing it.
4. box 2 work is done.
5. fabric quits at this point.

aravind:
https://github.com/fabric/fabric/pull/1471 works for me.

jsalva:
is there something holding up a merge of that PR?

goosemo:
@jsalva it's missing tests.
@aravind I'm not actually seeing that this feature works when specifically disabling it, I am testing in this way:


```
diff --git a/tests/test_parallel.py b/tests/test_parallel.py
index 3a8dbf3..7193ec6 100644
--- a/tests/test_parallel.py
+++ b/tests/test_parallel.py
@@ -51,11 +51,49 @@ class TestParallel(FabricTest):
                 run("ls /")
                 if env.host_string == host2:
                     raise OhNoesException
-
+
+            execute(mytask, hosts=[host1, host2])
+
+    @server(port=2200)
+    @server(port=2201)
+    @aborts
+    def test_parallel_quit_on_failure(self):
+        with settings(
+                hide('everything'),
+                parallel_exit_on_errors=True
+                ):
+            host1 = '127.0.0.1:2200'
+            host2 = '127.0.0.1:2201'
+
+            @parallel
+            def mytask():
+                run("ls /")
+                if env.host_string == host2:
[...]

PR #1471, opened 2016-06-02
Add an option to quit parallel execution in the event of an error

aravind:
Fixes issue #1469


fabric/state.py
```diff
@@ -224,6 +224,13 @@ def _rc_path():
         help="default to parallel execution method"
     ),
 
+    make_option('--quit-parallel-run-on-errors',
```
goosemo: I'd say have the flag be the same name as the var, so `--parallel-exit-on-errors`

fabric/job_queue.py
```diff
@@ -138,6 +140,11 @@ def _advance_the_queue():
 
         # Main loop!
         while not self._finished:
+            if self._abort:
```
goosemo: Didn't you want this to let the other running tasks complete their work?

tests/test_parallel.py
```diff
@@ -51,11 +51,33 @@ def mytask():
...
+            @parallel
+            def mytask():
+                run("ls /")
+                if env.host_string == host2:
+                    raise OhNoesException
+
+            try:
+                execute(mytask, hosts=[host1, host2])
+            except Exception:
+                pass
```
goosemo: Think you need to not have this `pass`, as you're getting a failed build with:

`AssertionError: test_parallel_quit_on_failure() did not raise SystemExit`

aravind: I will look into this somemore, but fab test on my machine passed.

fabric/job_queue.py
aravind: yeah, it might make sense to let the currently running jobs finish. If we want that behaviour, I don't even need to raise this exception. I can set queued to [] and let things end gracefully.

fabric/state.py
aravind: will change it to parallel-exit-on-errors.

aravind:
I couldn't figure out how to make a proper test case for the case where it is supposed to terminate after a few failures, so I removed my changes to tests/test_parallel.py.

fabric/state.py
```diff
@@ -224,6 +224,13 @@ def _rc_path():
         help="default to parallel execution method"
     ),
 
+    make_option('--parallel-exit-on-errors',
```
paulRbr: It would be nice to avec a short option version. `-E` seems to be free.

What do you think?

aravind: yeah, I will map it to -E, and submit a patch.

paulRbr: Thanks!

Em 30/07/2016 01:12, "Aravind Gottipati" <notifications@github.com>
escreveu:

paulRbr:
Any change to get this PR integrated in a feature release?

PR #1472, opened 2016-06-07
Remove the text about dependencies blocking the migration to Python 3

cclauss:
Remove `once our dependencies do` because all dependencies are already Python 3 compatible.

https://caniusepython3.com/project/Fabric


cclauss:
Bump.  It has been more than a year...

PR #1478, opened 2016-06-24
Add docstring to nested list

Nabellaleen:
Using a lot the modules and submodules, I like to use the nested list output.
The docstrings were not displayed with nested list output, so this is now the case.

Due to the recursive aspect of the _nested_list function, I've added a cache/memoization to some functions used in both nested and normal list output.
But I've activated this cache only on _nested_list, because _normal_list is called multiple times during an execution, with various context making the cache not work.


Nabellaleen:
I've let all my commit not squashed to help you understand some steps of the dev, but if you are ok with the PR, let me know, I'll squash its.

Nabellaleen:
Squashed !

Issue #1374, opened 2015-09-06
env.shell doesn't work properly on FreeBSD


hyperknot:
As based on the [FAQ](http://www.fabfile.org/faq.html#my-remote-system-doesn-t-have-bash-installed-by-default-do-i-need-to-install-bash), when using Fabric with a FreeBSD host, it needs 

`env.shell = "/bin/sh -c"`

However this doesn't work. Basic `run` commands are ok, but `append` and `echo $0` and `echo $SHELL`  are not.

code:
```
env.shell = "/bin/sh -c"

def get_shell():
    run('echo $0')
    run('echo $SHELL')
    append('a', 'b')
```

output:
```
[freebsd-virtualbox] run: echo $0
[freebsd-virtualbox] out: csh
[freebsd-virtualbox] out:

[freebsd-virtualbox] run: echo $SHELL
[freebsd-virtualbox] out: /bin/csh
[freebsd-virtualbox] out:

[freebsd-virtualbox] run: echo 'b' >> "$(echo a)"
[freebsd-virtualbox] out: Unmatched ".
[freebsd-virtualbox] out:


Fatal error: run() received nonzero return code 1 while executing!

Requested: echo 'b' >> "$(echo a)"
Executed: /bin/sh -c "echo 'b' >> \"\$(echo a)\""
```

The fix is to manually change the user's shell with chsh to sh or bash: `chsh -s /bin/sh` or `chsh -s /usr/local/bin/bash`

After this and a reconnect Fabric actually works, with the following output:
```
[freebsd-virtualbox] Executing task 'get_shell'
[freebsd-virtualbox] run: echo $0
[freebsd-virtualbox] out: /bin/sh
[freebsd-virtualbox] out:

[freebsd-virtualbox] run: echo $SHELL
[freebsd-virtualbox] out: /usr/local/bin/bash
[freebsd-virtualbox] out:

[freebsd-virtualbox] run: echo 'b' >> "$(echo a)"
```

I guess the question is that why doesn't env.shell change the shell when it's called on csh.

bitprophet:
Is this because FreeBSD's `/bin/sh` is actually cshell? Otherwise I don't quite get what's going on.

Also not sure what your question means exactly, are you asking why Fabric isn't trying to be more dynamic about which shell it selects?

hyperknot:
I'm reporting that Fabric doesn't work on a stock FreeBSD install, even with `env.shell = "/bin/sh -c"`.

[...]

PR #1483, opened 2016-07-07
Added to the FAQ more details about how to use csh.

mradziej:
There are multiple open issues in github about csh and quoting, also
on StackOverflow and else.

Background:

With normal settings (use_shell=True, shell_escape=True), a command that has
env.env_vars is wrapped and handed to the csh as:

/bin/bash -l -c "export FOO=\"bar\" && actual-command"

If in the csh setting backslash_quote is not set, csh does not handle the \"
as expected, but reads the command line as (one line per arg):

/bin/bash
-l
-c
export FOO=\
bar\
&& actual-command

... and chokes on the final "

### DIFF
--- a/sites/www/faq.rst
+++ b/sites/www/faq.rst
@@ -251,6 +251,12 @@ argument, like so::
 
 This has been shown to work on FreeBSD and may work on other systems as well.
 
+If the remote host uses ``csh`` for your login shell, fabric requires the shell variable
+backslash_quote to be set, or else the quoting will not work. If this is not the default,
+you probably want to add the following line to ``~/.cshrc``::
+
+    set backslash_quote
+
 
 I'm sometimes incorrectly asked for a passphrase instead of a password.
 =======================================================================


bitprophet:
Thanks for this!

You touched enough other outstanding issues (like #1374) that I'm probably going to try making a FBSD VM sometime so I can doublecheck the problem, but I'll be pretty happy if `backslash_quote` does make things a lot more compatible.

Wondering why this didn't come up prior - especially given the earlier FAQ content claiming that `/bin/sh -c` was sufficient to make things work. Could this be something that has changed in FreeBSD over the last 5-8 years? E.g.:

* Changing what `/bin/sh` points to, if it wasn't `csh` before;
* Upgrading to a new major version of `csh` that changed semantics/behavior/default config settings;
* Changes to the default systemwide shell config files (assuming FBSD does what many Linux distros do, and has default user dotfiles include a systemwide conf file);
* etc.

mradziej:
Hey, I'm glad if I was able to help out.

In my case, it's SUSE Linux SLES 11 patchlevel 4. I have no idea whether there is customization or whether it's the defaults that way. It's a confusing issue, and it might turn up only when there are environment variables. I did not test it on anything else, but from reading the documentation for csh it's quite obvious. In the bug reports you can also find the telltale sign message about the missing " ' ".

Fun fact: I requested to be changed to bash as login shell, but it took so long that I was really annoyed by the problem and took the time to read through. And now I'm on bash ;)

bitprophet:
In testing this out, it seems like the actual error message is simply `Unmatched ".` - going to add that to this update so it's got better SEO etc :)

PR #1484, opened 2016-07-08
Remove unused imports

ueg1990:


bitprophet:
Thanks! But this looks like it duplicates #1379, sorry :D (but it did prompt me to make sure that went into a milestone..)

Issue #465, opened 2011-10-31
Enhanced shell-like tilde expansion


bitprophet:
From a comment on #323, by @ngnpope:

> Would it be possible to make this behave more like shells do for tilde-expansion where the username after the tilde will expand to that user's home directory?  Some examples:

<table>
<tr><td>~</td><td>/home/user</td></tr>
<tr><td>~user</td><td>/home/user</td></tr>
<tr><td>~root</td><td>/root</td></tr>
<tr><td>~www-data</td><td>/var/www</td></tr>
</table>

PR #1486, opened 2016-07-17
Handling of remote ~username expansion

jeblundell:
Reproduction of bug (replace jim with a valid remote username):

> from fabric.api import *
> 
> def test():
>     put ("fabfile.py", "~jim/test.py")
>     get ("~jim/test.py", ".")

Output from bug:

```
[REMOTE] Executing task 'test'
[REMOTE] put: /tmp/fabric/fabfile.py -> /home/jimjim/test.py

Fatal error: put() encountered an exception while uploading '/tmp/fabric/fabfile.py'

Underlying exception:
    No such file

Aborting.
Disconnecting from REMOTE... done.
```

I'm proposing here to change tilde expansion to be done server-side. I've currently done this with a call to awk, but am open to changing it to another method. As documented in the function, simply doing a search for ~username and replacing with /home/username makes assumptions about where home directories are and so is not ideal.

I'm new to contributing to this repository, so do let me know if I've not met coding standards and I'll be happy to fix.


jeblundell:
I've just found that this is reported as issue #465 quite some time ago. Thoughts?

bitprophet:
Thanks for poking this. As noted in #465, a big problem here is having something cross-platform enough that it will work on e.g. Windows, BSD etc, as well as common Linux distros.

These days I'm more lenient and would say "it's ok to have a Linux-only addition, but it needs to degrade gracefully so it doesn't explode on other systems". So at least that'd mean adding some `warn_only=True` to your PR (then updating the logic so it deals well with an empty result).

Re: overall approach: I agree that scanning `/etc/passwd` is less-assumptive than assuming `/home`; I think in the past we've also considered using things like `echo $HOME`, which I'm sure has its own set of distinct tradeoffs, but might actually be better in some cases. (IIRC regular tilde expansion does something like that...might be good to look, consistency is a virtue.)

Finally, I'd feel more comfortable with this if there were some tests - best is probably to add something in `integration/` since this functionality is very "real world" focused, but we do have some mock-oriented tests in `tests/test_operations.py` that could be used as a guide for some tests there as well.

jeblundell:
Thanks for taking a look, much appreciated.

Ah, cross-platform. Hmm. Some additional thoughts
(1) ~username currently doesn't work, so if I break it then nobody will notice, so to an extent can do what I want (i.e. carry on)
(2) ~ is probably used by quite a lot of people potentially on non-Linux systems, so if I break it (even with careful testing) then it's a bit of a disaster

Do you think a sensible approach might be a compromise between existing system and my proposal where ~username is parsed using remote and ~ is parsed using old system? I can add a parameter to _remote_tilde_expand.

ploxiln:
Consider using `os.path.expanduser()`, which does something very similar, and attempts to handle Windows as well. https://docs.python.org/2/library/os.path.html#os.path.expanduser

ploxiln:
Oh, nevermind me, `os.path.expanduser()` would require python on the remote system, of course. My mistake, sorry. (Still, you might take some inspiration from how that claims to work.)

fabric/operations.py
```diff
@@ -244,6 +244,34 @@ def prompt(text, key=None, default='', validate=None):
...
+            expand_user = remote_path[1:]
+            remote_path = ''
+        else:
+            expand_user = remote_path[1:remote_path.index('/')]
+            remote_path = remote_path[remote_path.index('/')+1:]
+
+        if expand_user == '':
+            expand_user = env.sudo_user if use_sudo else env.user;
+
+        expanded = _run_command (command="awk -F: '$1 == \"%s\" {print $6}' /etc/passwd" % (expand_user), sudo=use_sudo, quiet=True)
```
ploxiln: I don't think you need sudo to read /etc/passwd, it is world-readable on all linux systems I've seen

ploxiln:
I think just running `echo ~user` in the shell on the remote is a decent way to do this.

It works for non-standard home directories on linux e.g. 

```
$ echo ~uucp
/var/spool/uucp
$ echo ~root
/root
```

It works on Mac OS, which does not put user accounts in /etc/passwd

```
$ whoami
pierce
$ grep pierce /etc/passwd
$ tail -n3 /etc/passwd
_ondemand:*:249:249:On Demand Resource Daemon:/var/db/ondemand:/usr/bin/false
_xserverdocs:*:251:251:OS X Server Documents Service:/var/empty:/usr/bin/false
_wwwproxy:*:252:252:WWW Proxy:/var/empty:/usr/bin/false
```

It does not depend on awk, only the standard shell.

I'm not sure what the situation is with windows, under what circumstances someone has an sshd running on windows and what shell is used for it. Keeping `ftp.normalize('.')` for the case of `~`, maybe just for windows, or if env.shell is not sh or bash, would make sense for keeping some compatibility for windows.

PR #1490, opened 2016-07-22
#1475: pass timeout to transport open_session

### LINKED ISSUE #1475: Pass a timeout value when requesting a new channel to the server
Hi,

Here: https://github.com/fabric/fabric/blob/master/fabric/state.py#L409 there is a no-arg call to open_session() 
and here: https://github.com/paramiko/paramiko/blob/master/paramiko/transport.py#L677 open_session() supports a keyword timeout parameter. 

open_session() calls open_channel() which in turn would wait for up to 1h before raising a connection error, 
see here: https://github.com/paramiko/paramiko/blob/master/paramiko/transport.py#L788
and here: https://github.com/paramiko/paramiko/blob/master/paramiko/transport.py#L828

Would it be possible to pass env.timeout or env.command_timeout in the call to open_session() to make the wait time configurable ?

jrmsgit:
Try to fix #1475 passing env.timeout to transport's open_session()

### DIFF
--- a/fabric/state.py
+++ b/fabric/state.py
@@ -407,7 +407,8 @@ connections = HostConnectionCache()
 
 
 def _open_session():
-    return connections[env.host_string].get_transport().open_session()
+    t = env.timeout
+    return connections[env.host_string].get_transport().open_session(timeout=t)
 
 
 def default_channel():


bitprophet:
Subsumed into maintainer change, will credit in changelog tho :)

PR #1495, opened 2016-07-31
Cross-platform support

Sahloul:
Windows can now run OpenSSH server, however, `test` is not provided in portable GNU executables.
As per issue #1496, `stat` would provide a better replacement, especially since it exits in *NIX and Windows GNU portable executables!

### DIFF
--- a/fabric/contrib/files.py
+++ b/fabric/contrib/files.py
@@ -28,11 +28,7 @@ def exists(path, use_sudo=False, verbose=False):
     """
     func = use_sudo and sudo or run
     with settings(hide('everything'), warn_only=True):
-        try:
-            result = func("ver") #is there better way to detect remote windows?
-            cmd = 'stat %s' % (_expand_path(path) if result.failed else path)
-        except:
-            cmd = 'test -e %s' % _expand_path(path)
+        cmd = 'stat %s' % (_expand_path(path) if func("ver").failed else path)
     # If verbose, run normally
     if verbose:
         with settings(warn_only=True):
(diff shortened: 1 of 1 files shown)


bitprophet:
@Sahloul notes in #1496 that `ver` is used here as a sentinel to determine remote platform. I don't know if it's the best way to do this offhand - is it available in all commonly found Windows versions out there? Are there any packages in popular Linux distros that might install a `ver` binary? etc.

Another random idea is to use the SFTP functionality to test file existence; though this has its own drawbacks (not all SSHDs have SFTP subsystem enabled; not usually possible to leverage sudo superpowers with SFTP; etc.)

Sahloul:
@bitprophet You are right! Depending on `ver` was not a good idea after all..
I have proposed a clean workaround for both remote file existence and path expansion compatibility.

Sahloul:
@bitprophet Mind to check what are the reasons for the build failure? I cannot understand its messages...

bitprophet:
@Sahloul the error is here: https://travis-ci.org/fabric/fabric/jobs/150021540#L1221 - looks like your docstring just isn't formatted as well as Sphinx wants, you need a code block or inline monospace or something. I can handle it whenever I merge, don't worry about it :) thanks!

Sahloul:
It was just the asterisk in *NIX ;)

bitprophet:
Originally cherry-picked this to 1.10 but then changed my mind; Github still isn't noticing though. Meh. It's in.

PR #1497, opened 2016-08-01
Fix a show_commands() function's issue:

ddcatgg:
Consider the task function's comment of a fabfile.py contains non ASCII chars, just like this:

```
# -*- coding: utf-8 -*-
from __future__ import unicode_literals
from fabric.api import task, run

@task
def test():
    """
    测试1234
    """
    run('env')
```

Assume it need to be executed through a parallel.py script which uses subprocess.Popen:

```
import subprocess

logfile = 'fabric.log'
logf = open(logfile, 'a')
args = ['fab', '-Hroot@192.168.100.18', 'test2']
p = subprocess.Popen(args, stdout=logf, stderr=logf)
p.wait()
```

The new process will throw a exception, look at the fabric.log:

```

Warning: Command(s) not found:
    test2

  File "c:\python27\lib\site-packages\fabric\main.py", line 729, in main
    show_commands(None, options.list_format, 1)
  File "c:\python27\lib\site-packages\fabric\main.py", line 592, in show_commands
    print("\n".join(list_commands(docstring, format)))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 31-32: ordinal not in range(128)
```

Because the comment in fabfile.py is unicode, while redirect to logf, it would be encoded to string using default encoding (ascii), and therefore failed.

To solve this issue, I modified the show_commands() function in fabric/main.py, add some code to change the output unicode to bytes explicitly.

Best wishes.


fabric/main.py
```diff
@@ -589,7 +589,12 @@ def update_output_levels(show, hide):
 
 
 def show_commands(docstring, format, code=0):
-    print("\n".join(list_commands(docstring, format)))
+    content = "\n".join(list_commands(docstring, format))
+    if isinstance(content, unicode):
+        import codecs, locale
```
ploxiln: I don't think it helps to import these standard library modules conditionally. Why not import them at the top of the file like usual?

ddcatgg: You are right, but it's about the grammatical style, does not matter ...

PR #1527, opened 2016-11-09
#1526-fix sed issues on MAC OS X

### LINKED ISSUE #1526: OS X Yosemite “sed: illegal option -- r” fatal error
I'm using fabric to deploy a system on MAC. However, I got the following issue.
> [localhost] out: sed: illegal option -- r
[localhost] out: usage: sed script [-Ealn] [-i extension] [file ...]
[localhost] out:        sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
[localhost] out:

> Fatal error: run() received nonzero return code 1 while executing!

The configuration of the machine:
MAC OS: OS X Yosemite
version 10.10.2
Fabric 1.10.2
Paramiko 1.15.2
Python 2.7.10

doflink:
fixes #1526 1526 
I have tested with python 2.7.10

### DIFF
--- a/fabric/contrib/files.py
+++ b/fabric/contrib/files.py
@@ -240,7 +240,7 @@ def sed(filename, before, after, limit='', use_sudo=False, backup='.bak',
 && cp -p %(filename)s %(filename)s%(backup)s \
 && mv %(tmp)s %(filename)s"""
     else:
-        context['extended_regex'] = '-E' if platform == 'Darwin' else '-r'
+        context['extended_regex'] = '-E' if 'Darwin' in platform else '-r'
         expr = r"sed -i%(backup)s %(extended_regex)s -e %(script)s %(filename)s"
     command = expr % context
     return func(command, shell=shell)


ploxiln:
```
    platform = run("uname")
    print("platform=%r" % platform)
```

shows exactly `platform='Darwin'` for me, on OS X 10.11.6, what does it show for you?

bitprophet:
Same here on Sierra (10.12.3). Not sure what else would be causing this PR to make the functionality work, though. Extra whitespace? We'll definitely need to see what @doflink's `uname` looks like :)

doflink:
Hi, 
when I convert  `platform`  to string and print out, it shows also `"Darwin"`. However, when I write it to a file, it shows `"^[(B^[[mDarwin"`. There is a special character in front of `"Darwin"` in my case.
Thanks.

bitprophet:
Likely some shell related artifact, since the value comes out of a `run('uname')`. Wonder if trying harder to avoid that in the `run()` (e.g. `use_shell=False` or similar; or stripping nonprinting characters; etc) is more appropriate than the `in` test; honestly not sure. It's not like a system with `Darwin` in its `uname` that _is not_ OS X is likely at all...

ploxiln:
It does look like something is adding some terminal escape codes - running without shell and/or pty is a good idea to try. (I'm pretty sure "uname" itself never does that, unless "uname" on your system is something besides the stock `/usr/bin/uname`).

```python
run("uname", shell=False, pty=False)
```

Maybe also make sure the output of `which uname` is `/usr/bin/uname`.

doflink:
I agree. Your solution is much better.

PR #1529, opened 2016-11-14
the proxy command set into env

kievdevops:
It is necessary to set proxy command via env

### DIFF
--- a/fabric/network.py
+++ b/fabric/network.py
@@ -83,7 +83,7 @@ def get_gateway(host, port, cache, replace=False):
     """
     from fabric.state import env, output
     sock = None
-    proxy_command = ssh_config().get('proxycommand', None)
+    proxy_command = getproxycommand() or ssh_config().get('proxycommand', None)
     if env.gateway:
         gateway = normalize_to_string(env.gateway)
         # ensure initial gateway connection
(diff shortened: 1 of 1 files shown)


jonaubf:
Very useful changes
Please, merge it ASAP

frooktz:
+1

Zlika:
This feature is a must have, because current version of fabric does not work with a ProxyCommand line in ~/.ssh/config when using a non standard SSH port (<> 22). I think the problem is located in paramiko (or it's the way fabric uses paramiko), in file https://github.com/paramiko/paramiko/blob/master/paramiko/config.py and in method _expand_variables: the port number defaults to 22, and then %p is always replaced by 22 in ProxyCommand.

Also, replace print "Proxy command must be string" by print("Proxy command must be string") to be Python3 compliant.

ploxiln:
It's probably better to make `ProxyCommand` in an ssh config work as expected, instead of adding a new `env` property (which overlaps a bit with `gateway`).

Currently paramiko only looks for the port in the config, so if you specified a different port in another way (e.g. the `--port` flag to fab) it does not get through to `_expand_variables()`. That's what should probably be fixed (and it will probably need some changes in paramiko, perhaps a way to `lookup()` without expanding variables yet).

PR #1548, opened 2017-01-06
README and Python 3.x support

neilsh:
Added info to README on Python 3.x support, to help avoid confusion for users.

(I'm not affiliated with Fabric3, just a user who ran into this confusion and would have found this information helpful)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,10 @@
 Fabric is a Python (2.5-2.7) library and command-line tool for
 streamlining the use of SSH for application deployment or systems
-administration tasks.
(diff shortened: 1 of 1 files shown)


bitprophet:
That fork is unofficial :( sorry! Fabric 2 will have Python 3 support.

herrbischoff:
@bitprophet: Thanks for letting us know. However, it's now been over a year and Fabric3 works with Python 2 and 3 while Fabric still only works with Python 2.

Is there anything in particular avoiding merging the changes back to this "official" version?

PR #1558, opened 2017-02-04
put() and get(): use random temp file when use_sudo=True

### LINKED ISSUE #1555: Race Condition in put when use_sudo=True
Abstract
-----------

There is a race in `put`, and the `use_sudo=True`, and writing a file with the same content from multiple connections concurrently.

Steps To Reproduce
---------------------------

Run this script concurrently:

```python
import os, itertools, StringIO

from fabric.api import task, run, put, settings

@task
def megaput():
	content = '0' * 1024

	with settings(host_string = '127.0.0.1', user = os.environ['USER']):
		run('mkdir -p /tmp/megaput')

		for i in itertools.count():
			put(StringIO.StringIO(content), '/tmp/megaput/file', use_sudo = True)
			print 'copied', i, 'times'
```

It seems that the name of temporary file created is a hash of the host string and the file content. When `put`ing files with the same content to the same host, there is a race for the `mv` operation, and with non-zero probability only the first operation will succeed.

Adding a random string to https://github.com/fabric/fabric/blob/master/fabric/sftp.py#L164 seems to make this script work without problems (e.g. `hasher.update(uuid.uuid4().hex)`).

ploxiln:
This was using a hash of the host and remote file path,
and could result in an unnecessary failure if there were
multiple get() and/or put() of the same file path on the same host
simultaneously.

fixes #1555

### DIFF
--- a/fabric/sftp.py
+++ b/fabric/sftp.py
@@ -1,10 +1,10 @@
 from __future__ import with_statement
 
-import hashlib
 import os
 import posixpath
 import stat
 import re
+import uuid
 from fnmatch import filter as fnfilter
 
 from fabric.state import output, connections, env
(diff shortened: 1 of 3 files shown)


ploxiln:
Failed tests are an example of why "unit" tests are dumb - they test implementation details, not correct results. Will look into it ...

bitprophet:
To save me from making a dumb mistake as with that other ticket, can you rehome this to 1.12 when you have a sec? Thanks!

ploxiln:
changing base forward is easier :)

ploxiln:
Test failures are the `contains()` `case_sensitive` ones - I'll rebase after #1596 is merged and they'll pass.

PR #1571, opened 2017-03-01
contrib.sed(): avoid shell for uname

### LINKED ISSUE #1526: OS X Yosemite “sed: illegal option -- r” fatal error

ploxiln:
Some (uncommon) systems can end up with shell escapes in the output
of run("uname"), so don't use shell (or pty), they're not needed anyway.

fixes #1526 / #1527

Figured this was another very easy little "fix". I'm happy to re-target to whatever branch is desired.

There is an integration test for `contrib.sed()` (already), which should be a reasonable sanity-test that I didn't break it...

### DIFF
--- a/fabric/contrib/files.py
+++ b/fabric/contrib/files.py
@@ -227,7 +227,7 @@ def sed(filename, before, after, limit='', use_sudo=False, backup='.bak',
     # Test the OS because of differences between sed versions
 
     with hide('running', 'stdout'):
-        platform = run("uname")
+        platform = run("uname", shell=False, pty=False)
     if platform in ('NetBSD', 'OpenBSD', 'QNX'):
         # Attempt to protect against failures/collisions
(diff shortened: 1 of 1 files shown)


bitprophet:
+1 on disabling shell/pty for this sort of background call. Thanks :)

PR #1572, opened 2017-03-01
Try to connect multiple times before prompt for password

GregorioDiStefano:
while attempting to SSH into a device that is booting up, a password prompt will incorrectly be prompted (in reality there is no password set for SSH on the remote host) to the user. If abort_on_prompts is set, SystemExit is thrown. Currently, we are catching the BaseException, and trying again.

My fix tries to connect multiple times before asking for a password. 

Please let me know if you need more info


bitprophet:
Hrm. I get the gist and it seems reasonable for the case you mention, but I wonder how it might screw up other cases. E.g. in the case where a password is required and just isn't being set correctly, this will cause `connection_attempts` failed auths, which could trigger a lockout if `connection_attempts` is too high.

That said, that value defaults to `1`, so it would really only be problematic if one both has it explicitly set to a higher value _and_ also managed to goof up password configuration. Still possible, but not as likely.

----

The "real" fix for this, I suspect, is for Paramiko to be better about exception-raising (A long standing issue) so that Fabric doesn't have to assume it's facing a bad password when it gets to this point. That's an issue for that repo and/or fab v2 though :(

----

What else could we do here?

- Make this an opt-in behavior...that'd make it safer to merge, esp at this point in v1's lifetime I'm leery of behavioral changes.
- Find a better way to detect your particular situation - an sshd that is up enough to handshake but which will cause some exceptions to be raised from Paramiko? (Maybe the full tracebacks you're getting might be helpful; there might be some smaller change we could make on the Paramiko end to make this easier to detect.)

GregorioDiStefano:
> Find a better way to detect your particular situation - an sshd that is up enough to handshake but which will cause some exceptions to be raised from Paramiko? (Maybe the full tracebacks you're getting might be helpful; there might be some smaller change we could make on the Paramiko end to make this easier to detect.)

This sounds like a reasonable approach. I can take a look and see what's happening

GregorioDiStefano:
Interesting:

```
DEBUG:paramiko.transport:starting thread (client mode): 0x9d46b190L
DEBUG:paramiko.transport:Local version/idstring: SSH-2.0-paramiko_2.1.1
ERROR:paramiko.transport:Exception: Error reading SSH protocol banner
ERROR:paramiko.transport:Traceback (most recent call last):
ERROR:paramiko.transport:  File "/usr/local/lib/python2.7/dist-packages/paramiko/transport.py", line 1744, in run
ERROR:paramiko.transport:    self._check_banner()
ERROR:paramiko.transport:  File "/usr/local/lib/python2.7/dist-packages/paramiko/transport.py", line 1892, in _check_banner
ERROR:paramiko.transport:    raise SSHException('Error reading SSH protocol banner' + str(e))
ERROR:paramiko.transport:SSHException: Error reading SSH protocol banner
ERROR:paramiko.transport:
DEBUG:paramiko.transport:starting thread (client mode): 0x9c33ddd0L
DEBUG:paramiko.transport:Local version/idstring: SSH-2.0-paramiko_2.1.1
ERROR:paramiko.transport:Exception: Error reading SSH protocol banner[Errno 104] Connection reset by peer
ERROR:paramiko.transport:Traceback (most recent call last):
ERROR:paramiko.transport:  File "/usr/local/lib/python2.7/dist-packages/paramiko/transport.py", line 1744, in run
ERROR:paramiko.transport:    self._check_banner()
ERROR:paramiko.transport:  File "/usr/local/lib/python2.7/dist-packages/paramiko/transport.py", line 1892, in _check_banner
ERROR:paramiko.transport:    raise SSHException('Error reading SSH protocol banner' + str(e))
ERROR:paramiko.transport:SSHException: Error reading SSH protocol banner[Errno 104] Connection reset by peer
ERROR:paramiko.transport:
```

is repeated many times, until eventually I get the prompt:

Fatal error: Needed to prompt for a connection or sudo password (host: 172.21.0.5:8822), but abort-on-prompts was set to True

Aborting.

bitprophet:
Huh. My low-level networking isn't amazing so I can't think offhand why an ssh connection would be getting `Connection reset by peer` instead of simply `Connection refused`.

Also funny that you're getting that SSHException multiple times, is that because of your `except BaseException` retry loop as well? Seems like code above your addition ought to be raising `NetworkError` after `connection_attempts` retries, if it's getting the protocol banner error.

EDIT: ooh, no, that code is too limiting, it uses `==` instead of `startswith` and the protocol banner message in your case is sometimes slapping in the reset-by-peer message. While this would not truly solve your issue (you'd just be getting the `NetworkError` "too early" instead of the prompt-driven SystemExit "too early") it's an interesting wrinkle.

Issue #1574, opened 2017-03-04
upload_project fails


aidanmelen:
I am simply trying to copy a directory from my local machine to a virtual machine using **fabric.contrib.project.upload_project** feature.

my environment is:
python==2.7.12
fabric==1.13.1

here is the directory structure :
.
├── .vagrant
│      └── ...
├── Vagrantfile
├── .local_dir
│   └── file.txt
└── upload_project_test.py

**upload_project_test.py** contains the following code

```
from fabric.contrib.project import upload_project
from fabric.state import env
from os import path, getcwd, sep
from fabric.context_managers import lcd
from fabric.operations import local

project_root_path = getcwd()
ssh_config = None

with lcd(path.join(project_root_path)):
    if 'running' in local('vagrant status', capture=True):
        ssh_config_lines = local('vagrant ssh-config', capture=True)
        ssh_config = dict(line.split() for line in ssh_config_lines.split('\n'))

env.user = ssh_config['User']
env.key_filename = ssh_config['IdentityFile']
env.host_string = '{user}@{host}:{port}'.format(
                                   user=ssh_config['User'],
                                   host=ssh_config['HostName'],
                                   port=ssh_config['Port'])

local_dir = path.join(project_root_path, '.local_dir')
remote_dir = path.join(sep)    # vm root

upload_project(local_dir=local_dir, remote_dir=remote_dir, use_sudo=True)
```

**this will fail because .local dir is hidden.**

```
[localhost] local: vagrant status
[localhost] local: vagrant ssh-config
[localhost] local: tar -czf /var/folders/8z/dm68k1vd1zxf73slr540s3wc0000gn/T/tmpMGJuIH/local_dir.tar.gz -C /Users/aidanmelen/Sampleminded/infrastructure/python_tools/tastetest/test local_dir
tar: local_dir: Cannot stat: No such file or directory
tar: Error exit delayed from previous errors.

Fatal error: local() encountered an error (return code 1) while executing 'tar -czf /var/folders/8z/dm68k1vd1zxf73slr540s3wc0000gn/T/tmpMGJuIH/local_dir.tar.gz -C /Users/aidanmelen/Sampleminded/infrastructure/python_tools/tastetest/test local_dir'

Aborting.
[localhost] local: rm -rf /var/folders/8z/dm68k1vd1zxf73slr540s3wc0000gn/T/tmpMGJuIH
```

it will work properly if '.local_dir; is not hidden, such as 'local_dir'

[...]

PR #1576, opened 2017-03-07
fix upload_project() for local_dir with no path separator

ploxiln:
A tiny fix for a contrib function. Bug identified in #1574 

My thinking is: fabric probably does not want to maintain these contrib functions, and will remove them in 2.0+, and I personally don't use them. But if there is a very easy fix for a very simple case, might as well put it in and avoid some bug reports.

The root of the problem here is `os.path.split()` can return a blank parent path, which doesn't work here.

```
>>> os.path.split("./testdir")
('.', 'testdir')
>>> os.path.split("testdir")
('', 'testdir')
```

In addition to the (trivial) unit test, I did a manual test without and with the fix.

before:

```
$ fab -H ubuntu@devbox test
[ubuntu@devbox] Executing task 'test'
[localhost] local: tar -czf /var/folders/ny/tsl8z9qx453630cxnzw0_dj80000gn/T/tmp52uz2D/testdir.tar.gz -C  testdir
tar: no files or directories specified

Fatal error: local() encountered an error (return code 1) while executing 'tar -czf /var/folders/ny/tsl8z9qx453630cxnzw0_dj80000gn/T/tmp52uz2D/testdir.tar.gz -C  testdir'
```

after:

```
$ fab -H ubuntu@devbox test
[ubuntu@devbox] Executing task 'test'
[localhost] local: tar -czf /var/folders/ny/tsl8z9qx453630cxnzw0_dj80000gn/T/tmpatuzuo/testdir.tar.gz -C . testdir
[ubuntu@devbox] put: /var/folders/ny/tsl8z9qx453630cxnzw0_dj80000gn/T/tmpatuzuo/testdir.tar.gz -> /tmp/testdir.tar.gz
[ubuntu@devbox] run: tar -xzf testdir.tar.gz
[ubuntu@devbox] run: rm -f testdir.tar.gz
[localhost] local: rm -rf /var/folders/ny/tsl8z9qx453630cxnzw0_dj80000gn/T/tmpatuzuo
```

I can re-base this on an older branch (e.g. 1.11) and add a changelog entry if desired.

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -104,7 +104,23 @@ class UploadProjectTestCase(unittest.TestCase):
         # local() is called more than once so we need an extra next_call()
         # otherwise fudge compares the args to the last call to local()
         self.fake_local.with_args(
-            arg.endswith("-C %s %s" % os.path.split(project_path))
+            arg.endswith("-C path/to/my project")
+        ).next_call()
+
+        # Exercise
+        project.upload_project(local_dir=project_path)
+
+
+    @fudge.with_fakes
+    def test_path_to_local_project_no_separator(self):
+        """Local folder can have no path separator (in current directory)."""
+
+        project_path = "testpath"
+
+        # local() is called more than once so we need an extra next_call()
+        # otherwise fudge compares the args to the last call to local()
+        self.fake_local.with_args(
+            arg.endswith("-C . testpath")
         ).next_call()
 
         # Exercise
--- a/fabric/contrib/project.py
+++ b/fabric/contrib/project.py
@@ -186,6 +186,7 @@ def upload_project(local_dir=None, remote_dir="", use_sudo=False):
     local_dir = local_dir.rstrip(os.sep)
 
     local_path, local_name = os.path.split(local_dir)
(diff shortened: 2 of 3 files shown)


tests/test_project.py
```diff
@@ -104,7 +104,23 @@ def test_path_to_local_project_can_be_specified(self):
         # local() is called more than once so we need an extra next_call()
         # otherwise fudge compares the args to the last call to local()
         self.fake_local.with_args(
-            arg.endswith("-C %s %s" % os.path.split(project_path))
+            arg.endswith("-C path/to/my project")
```
ploxiln: tangential commentary: another unit-test pitfall is duplicating the code being tested in the test. In this case it works, but if it didn't, it would still pass ...

ploxiln:
changelog added

ploxiln:
re-targetted to 1.10

bitprophet:
Thanks for this. Merged in 1.12 (yea...figures, once someone targets 1.10 like I keep asking, I shrink the support window 🙃 ) and up.

ploxiln:
no problem, thanks!

PR #1581, opened 2017-04-03
Adds --use-default-prompts flag

viniciusd:
This flag is used to prioritize default values over prompting the user for an input.
Sometimes, we just want to automatically reply all the prompts and get the task done without user interference. Prompting should be switchable (on/offf).
For example, developers may answer the prompts for customizing their run while, when running in a CI cycle, the prompts should get the default value.


viniciusd:
Doesn't look like the failure has something to do with this PR

viniciusd:
Any feedback on why so I can fix it, @bitprophet ? It seems this merge request wasn't properly done because it shows like it's adding a bunch of commits, I will look into it.
I can also add the related changelog/docs.

PR #1586, opened 2017-04-07
fixes #1341: clean up temp file with sudo uploads if destination path…

### LINKED ISSUE #1341: Security problem: upload_template () ==> world readable file in the home folder
I wast running the upload_template() function from the fabric.contrib.files package and found an unexpected behavior.
If the "destination" parameter points to an invalid path then you end up with a world-readable file in your home folder:

-rw-r--r--  1 me users  623 Jun 11 12:02 dbdc6b14139b9aaf18cfcd2cb1244440dbf08136

If the file happens to contain a password there is a chance of the somebody reading it.

UkuLoskit:
… is invalid

### DIFF
--- a/fabric/sftp.py
+++ b/fabric/sftp.py
@@ -287,7 +287,11 @@ class SFTP(object):
             # Temporarily nuke 'cwd' so sudo() doesn't "cd" its mv command.
             # (The target path has already been cwd-ified elsewhere.)
             with settings(hide('everything'), cwd=""):
-                sudo("mv \"%s\" \"%s\"" % (remote_path, target_path))
+                try:
+                    sudo("mv \"%s\" \"%s\"" % (remote_path, target_path))
+                except BaseException:
+                    sudo("rm -f \"%s\"" % remote_path)
+                    raise
             # Revert to original remote_path for return value's sake
             remote_path = target_path
         return remote_path


UkuLoskit:
Any comments on this one?

bitprophet:
(for easier forward referencing: #1341)

Not sure how this got past me originally...looks all right at a glance, thanks!

Wondering if `try/finally` makes more sense here, it's normally what one wants to do in this sort of spot (i.e. my mental alarm bells went off at `except BaseException`), and since one uses `rm -f`, it will not error even if the `mv` succeeded.

Only downside is in the success case it's one extra shell command that is not strictly necessary, but I don't know if that bothers me a ton. Readability vs efficiency, I guess.

UkuLoskit:
Thanks for the quick feedback, @bitprophet. This was exactly my line of reasoning that calling `rm` always would be a bit wasteful, for example in a scenario where you are uploading a large file tree. 

I could still change it, if you'd like.

bitprophet:
I think I'm okay with it as-is for now; as much as I enjoy beating people over the head with the readability clause, an extra almost-never-useful shell session inside something this frequently used & which would execute unconditionally, just leaves a bad taste.

bitprophet:
Merged into 1.14 branch.

PR #1590, opened 2017-04-18
Use `python -m fabric` rather than `fab` in tests.

BenSturmfels:
The `test_abort_message_only_printed_once` test needs Fabric already installed to run, since the `fab` executable is created only during install. This wouldn't ordinarily be a problem, but when packaging Fabric for a system like GNU Guix, we'd like to be able to run the tests before installing.

This change switches this single use of `fab` to the equivalent `python -m fabric`.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -93,7 +93,7 @@ def test_abort_message_only_printed_once():
     # perform when they are allowed to bubble all the way to the top. So, we
     # invoke a subprocess and look at its stderr instead.
     with quiet():
-        result = local("fab -f tests/support/aborts.py kaboom", capture=True)
+        result = local("python -m fabric -f tests/support/aborts.py kaboom", capture=True)
     # When error in #1318 is present, this has an extra "It burns!" at end of
     # stderr string.
     eq_(result.stderr, "Fatal error: It burns!\n\nAborting.")


BenSturmfels:
Thanks also for a brilliant tool which I use nearly every day for work! :)

bitprophet:
I'm sure I'm missing something silly but how are you able to use `-m fabric` (implying `fabric` is on `sys.path` and thus at least partially installed) yet not `fab` (implying `setup.py install` has not been run or entrypoints are otherwise being skipped)?

That said I have no actual problem with this, just want to understand the background more. Thanks!

BenSturmfels:
The `-m fabric` works because the [current directory is added](https://docs.python.org/2/library/sys.html#sys.path) to `sys.path`, and the top level of the source tree where the tests are run from contains the module `fabric` (a directory called "fabric" that contains an "__init__.py). It sounds like you're overthinking it, but let me know if I'm missing something silly myself.

BenSturmfels:
Sorry just realised that I didn't understand this as well as I thought - it's because it has an `__init__.py` *and* an `__main__.py` that it can be run with `-m fabric`.

bitprophet:
Ah, gotcha - forgot about the CWD-on-path angle, that makes sense now. Thanks!

bitprophet:
Realized I never asked _how_ exactly you run the tests pre-install...while I don't use it often I suspected perhaps it was via `python setup.py test` (which appears to install test deps, and run `egg_info` and `build_ext` for Fabric itself, but not any actual installation).

Ran that in an env w/ Fabric uninstalled, and yup, I get a kaboom on the test(s) using the helper being patched.

Cherry-picked the patch and...hm, still doesn't work, I now get `fabric is a package and cannot be directly executed`; confirmed I get this by hand too with e.g. `python -m fabric -l`. Ah - this seems to be a Python 2.6 (my default dev version) vs 2.7 issue, works fine under 2.7. Cool.

bitprophet:
Cool, looks like we can simply explicitly request `fabric.__main__` and that works on both interpreter versions. Crisis averted 👍

PR #1593, opened 2017-04-21
Add missing "c." in upgrading.rst

RedKrieg:
These 4 run() calls should be method attributes of the "c" object in this task.

### DIFF
--- a/sites/docs/upgrading.rst
+++ b/sites/docs/upgrading.rst
@@ -272,8 +272,8 @@ expect it will be soon. For now we fall back to command chaining with ``&&``.
         if not c.run("test -d {}".format(code_dir), warn=True):
             cmd = "git clone user@vcshost:/path/to/repo/.git {}"
             c.run(cmd.format(code_dir))
-        run("cd {} && git pull".format(code_dir))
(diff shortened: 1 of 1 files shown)


bitprophet:
Boy, that CodeCov integration is ginormous. Will have to see if it's that huge every changeset, I might prefer the old just-a-comment style integration instead...

Thanks for the catch!

ploxiln:
My 2c: I prefer the coverage results just being on the commit (click the green checkmark to see them, click details for details). Otherwise, if someone pushes up fixup or in-progress commits one at a time, you get a lot of redundant notifications of comments from the coverage service.

bitprophet:
I'd honestly forgotten all about that checkmark, I see it now. Probably agree.

Does seem codecov offers a number of options around how it comments, what is in the comment, and whether it comments at all: https://docs.codecov.io/docs/pull-request-comments

At the very least it seems like it should not be spammy like it used to be, I assume that is due to using the newer integration style. Claims its default is to edit a single comment anytime the PR is updated, which won't generate new notifications.

Even so, as noted, I'm not sure I love the comment at all. I'll keep an eye on this (will be hard not to...) and modify the codecov.yml appropriately!

Issue #1594, opened 2017-04-22
v2-compatible 'roles' or similar


bitprophet:
## Synopsis

At time of writing, the v2 branch has a `Group` class that should be capable of serving as the units formerly known as 'roles', aka "a bunch of hosts to do stuff with/on".

However, there's no specific way of organizing or labeling `Group` objects yet; it's "done" enough for the pure API use case of advanced users who want to roll their own specific way of creating them, but lacks anything for CLI-oriented users or intermediate folks who want something frameworky to build around.

## Background

In v1, roles were effectively a single flat namespace mapping simple string labels to what would be Groups in v2, and they could be selected on the CLI at runtime (`fab --roles=web,db`) and/or registered as default targets for tasks (`@task('db') \n def migrate():`), much like hosts.

Users defined them in `env.roledefs`, a simple dict; any intermediate to advanced functionality revolved around modifying it, usually at runtime (via pre-task or subroutine), sometimes at module load time.

## Specific use cases / needs / subfeatures

- Basic, naive mapping for use/reference anywhere else in the system: put in a name, get back some iterable of `Group`s and/or `Connection`s.
  - Aliasing often wants to go along with that, so e.g. a `Lexicon` instead of a `dict`.
  - Even deeper constructs, such as 'bundling', e.g. you have direct mappings named `db`, `web`, `lb`, but then a 2nd-tier name called `prod` that is always the union of the other three. I forget if I added that to `Lexicon` yet. Possible there's other map subclasses out there that already do it too.
  - Additionally/alternately, things like globbing or other string syntaxes, though I personally would prefer to leverage the fact that Python is not "stringly typed"...
- Target selection on the CLI, globally and/or per-task
  - An extension of Invoke's CLI system to account for "flags that all tasks get on top of what they define" may be useful or required for this. Which falls firmly into pyinvoke/invoke#205 territory, in fact, so that just got higher priority than it already was (which was pretty high.)
- Ditto task-level defaults
  - Though task-level target defaults really want to be any of: connection, connections, group obj, group objs, or name evaluating _to_ group objs (that last is the only thing that directly pertains to this ticket, arguably)
- Ditto collection-level defaults (NEW in v2!)
  - I.e. "all tasks in <submodule> default to running against the `db` role"
  - Same deal as previous point - this default wants to allow a number of different values, not just a string key.

## Potential solutions

max-arnold:
From the mailing list:

[...]

PR #1595, opened 2017-04-23
Implement execute() for Groups

RedKrieg:
This implements the execute method of SerialGroup and ThreadingGroup.

I only wrote one test for now, as I'm not sure this implementation is the best approach and wanted feedback before I continued with reimplementing things like the executor to use these functions.

This implementation modifies each `connection`'s config to include those config variables normally passed as keyword arguments to `connection.run()`.  This gives us keyword control and lets the task's internal logic override us as necessary.  The cloned config is restored after execution (not sure this is the right approach here either).

I needed this functionality to begin implementing fabric/fabric#1594

### DIFF
--- a/fabric/group.py
+++ b/fabric/group.py
@@ -195,7 +195,6 @@ class ThreadingGroup(Group):
         results = GroupResult()
         queue = Queue()
         threads = []
-        old_configs = {}
         for cxn in self:
             my_kwargs = dict(
                 cxn=cxn,
--- a/fabric/group.py
+++ b/fabric/group.py
@@ -1,4 +1,5 @@
 from invoke.vendor.six.moves.queue import Queue
+from invoke.vendor.six import iteritems
 
 from invoke.util import ExceptionHandlingThread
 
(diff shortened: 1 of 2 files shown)


bitprophet:
Yea OK codecov's big comments need to go...doing that now heh.

Re: this PR itself: apologies, but I'm actually not at all sure `Group.execute` wants to be anything but a thin wrapper around an implicit `Executor`, much like how `invoke.run` is a thin wrapper around an implicit `Runner`.

This area (the overlap between Group and Executor) is not yet well defined, it needs at least some thinkin' before we start implementing it for reals. But I don't want "how to run a task" to end up split up or duplicated between those two classes, and since Executor definitely needs that management at the Invoke level (and for a bevy of other things not yet implemented, like deeper dependency management and non-Fabric-oriented task parameterization) it needs to be "the place" for those concerns, I think.

Now that I've started processing these thoughts I'll see if I can put something more firm down in at least a DDD fashion if not actual code.

bitprophet:
Some related tickets and labels that already exist:

- pyinvoke/invoke#170
- https://github.com/pyinvoke/invoke/labels/Task%20execution

breisig:
Can we get this functionality added and merged ASAP?

RedKrieg:
@breisig You're welcome to maintain this functionality in your own fork using my patches.  It is both rude and unreasonable to make demands of an open source project for which you pay nothing.  If you want this functionality ASAP, implement it yourself in a way compatible with the project vision (discussed in this thread) and submit a pull request.

Issue #1598, opened 2017-04-28
fabric.contrib.files.append does not handle input with newlines correctly


erik-cliqz:
In Fabric 1.13.2, the following code produces unexpected results
```
from fabric.contrib.files import append
append('file.txt', 'a') # appended                                                                                                                       
append('file.txt', 'b') # appended                                                                                                                       
append('file.txt', 'c\n') # not appended                                                                                                                 
append('file.txt', 'a\nd') # not appended
```
The resulting `file.txt` has the following content
```
a
b
```
This behaviour is not present in Fabric 1.13.1. It looks like `_escape_for_regex` is not properly escaping newlines.

ploxiln:
Likely caused by #1559 (which was applied to the 1.12 branch and later). I'll look into it ...

ploxiln:
The problem is with how egrep handles newline characters.

```
>> cat test.txt
ab
cd
>> egrep 'ab' test.txt
ab
>> egrep 'f' test.txt
>> echo $'^c\n$'
^c
$
>> egrep $'^c\n$' test.txt
ab
cd
>> egrep $'^f\n$' test.txt
ab
cd
>> egrep $'^f\nb$' test.txt
ab
```

`contrib.append()` is described as:

> Append string (or list of strings) text to filename.

> If text is already found in filename, the append is not run, and None is returned immediately.

grep/egrep is also explicitly about matching *lines*. I don't think there's any way to do multiple-line patterns with egrep. So, it seems that newline characters in the egrep pattern are considered to separate multiple patterns, any of which can match. So "\n$" matches every line, and "anything\n$" also matches every line. So `append()` always thinks the line is already present.

How did it work before? Newlines were effectively removed from the line when it was being checked (as a pattern). Thus typically no match, so it was appended. For example,

```
append('file.txt', 'a\nd')
```

was run by Fabric 1.13.1 as:

```
egrep "^a\
d$" "$(echo file.txt)"
```

In the shell, backslash-newline means to *ignore* the newline. So this was equivalent to:

```
egrep "^ad$" "file.txt"
```

So if you ran `append()` twice, it would append twice, which the documentation said it should not do, it should only append once. The second time it would not find what it had just appended.

Conclusions:

  * `contrib.files.append()` does not handle newline characters correctly. It never has, but it changed from "usually erroneously append" to "usually erroneously don't append". The documentation does say it's for lines, after all ... but it does "break" some backwards compatibility ... but then again, it's "contrib" ...

[...]

PR #1599, opened 2017-04-28
contrib.files: adjust _escape_for_regex() newline handling

ploxiln:
Remove newlines from the pattern completely. This is what the
original _escape_for_regex() effectively did. There is no way
to get egrep to handle newlines the desired way, so might as well
remain consistent with the original behavior regarding newlines.

See longer explanation in #1598

### DIFF
--- a/fabric/contrib/files.py
+++ b/fabric/contrib/files.py
@@ -433,7 +433,15 @@ def _escape_for_regex(text):
     re_chars = []
     sh_chars = []
 
+    # Removes newline characters.
+    # egrep does not support multi-line patterns, so they should not be used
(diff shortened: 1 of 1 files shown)


ploxiln:
whether it is a good idea to merge this change, I leave up to the judgement of @bitprophet

bitprophet:
We've clearly already broken a few eggs with the original change, might as well roll forward with it and try making an omelet. If it continues to feel unstable after this + a few more patches, we can undo all the changes and e.g. stuff them into a feature release instead (or just keep them out, tho I'd rather not. eh.)

ploxiln:
friendly ping

PR #1606, opened 2017-05-22
Python 3.6 compatibility; tested with `fab -h` only.

sanscore:
Quite a few syntactic changes, but functionally speaking, there's potentially a few. Here's a couple worth pointing out.

* Reimplementation of `nested`.
  * https://github.com/sanscore/fabric/blob/71f35201bebeebf4a98b4547b68730d1b53b3875/fabric/context_managers.py#L41-L54
* `raise_if_needed` in `thread_handling.py` uses `raise e[0], e[1], e[2]` which no longer supported. I've never seen `raise` used in that way, so I was unsure of what the modern equivalent is, so I just logged all the values and used a bare `raise` instead.
  * https://github.com/sanscore/fabric/blob/71f35201bebeebf4a98b4547b68730d1b53b3875/fabric/thread_handling.py#L22-L28

It'll be very interesting to see what TravisCI thinks because I cannot get Python2.5 to work on my local machine; `fab -h` works on py26, py27 and py36 without any problems.

### DIFF
--- a/fabric/context_managers.py
+++ b/fabric/context_managers.py
@@ -33,9 +33,25 @@ Context managers for use with the ``with`` statement.
     ``nested`` itself -- see its API doc for details.
 """
 
-from contextlib import contextmanager, nested
 import socket
 import select
+from contextlib import contextmanager
+
+# http://stackoverflow.com/a/39158985/185101
+try:
+    from contextlib import nested
+except ImportError:
+    from contextlib import ExitStack, contextmanager
+
+    @contextmanager
+    def nested(*contexts):
+        """
+        Reimplementation of nested in python 3.
+        """
+        with ExitStack() as stack:
+            for ctx in contexts:
+                stack.enter_context(ctx)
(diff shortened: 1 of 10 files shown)


bitprophet:
Travis doesn't even have 2.5 anymore unfortunately :(

I'm actually going to just-close this per our discussions re: fab 1 vs fab 2, but feel free to keep updating it for academic purposes. If you somehow find the holy grail of a 2.5->3.x compatible syntax...who knows.

sanscore:
Sounds good. I'll hook up my personal fork to my TravisCI account and report back.

apitop:
Hi,
This PR seems to solve a very old issue sinon later 2.8.1 !!! and we are many to be locked on this old issue because of that (see this discussion : [https://github.com/fabric/fabric/issues/2182](https://github.com/fabric/fabric/issues/2182) )
As this seems to some this major issue, could you import it in the next release of Paramiko ?
Regards

PR #1609, opened 2017-05-24
Issue 1493

### LINKED ISSUE #1493: Allow overriding default 5s post-'reboot' sleep in operations.reboot()
In some cases, a host may not stop network service in 5 seconds, such as many services should be shutdown,  but the reboot api try to reconnect the host,  then it report the host reboot success in wrong.

`
(sudo if use_sudo else run)(command)
        # Try to make sure we don't slip in before pre-reboot lockdown
        time.sleep(5)
`

mikebarkas:
Patch for adding `timeout` kwarg to the reboot function for sleep timeout.

Closes #1493


bitprophet:
Thanks for this! #1493 is in the 1.14 milestone so I'll hopefully get to just-merge this when I go through that list.

PR #1615, opened 2017-05-31
contrib.files: get rid of _expand_path() and is_win()

### LINKED ISSUE #367: files.append not detecting lines present if ~ used

ploxiln:
They're not needed. If the filename is just put into the command
string unquoted, it will be expanded by the shell.

(If ` "$(echo %s)" ` will work in the command string, then just ` %s ` will work the same. As a bonus, it won't need special handling for windows.)

My original motivation can be seen in #1614 - `is_win()` is annoying:

```
$ fab --show=debug -H testdeploy11.ec2.st-av.net test
...
[testdeploy11.ec2.st-av.net] Executing task 'test'
[testdeploy11.ec2.st-av.net] run: /bin/bash -l -c "echo \"Will you echo quotation marks\""
[testdeploy11.ec2.st-av.net] run: /bin/bash -l -c "stat \"\$(echo test.txt)\""
[testdeploy11.ec2.st-av.net] run: /bin/bash -l -c "echo \"Will you echo quotation marks\""
[testdeploy11.ec2.st-av.net] run: egrep "^127\\.0\\.0\\.1 db$" "$(echo test.txt)"
[testdeploy11.ec2.st-av.net] run: /bin/bash -l -c "echo \"Will you echo quotation marks\""
[testdeploy11.ec2.st-av.net] run: /bin/bash -l -c "stat \"\$(echo test.txt)\""
[testdeploy11.ec2.st-av.net] run: /bin/bash -l -c "echo \"Will you echo quotation marks\""
[testdeploy11.ec2.st-av.net] run: egrep "^127\\.0\\.0\\.1 redis$" "$(echo test.txt)"
```

That PR added a cache for this. But then I realized that none of this is needed at all.

It looks like `_expand_path()` was added in c6a458526c6dad9447c17c3e03e94233a791a715 to fix #367 - but just removing the quotes around the filename argument to egrep would have sufficed.

```diff
@@ -314,7 +314,7 @@ def contains(filename, text, exact=False, use_sudo=False, escape=True,
     with settings(hide('everything'), warn_only=True):
-        egrep_cmd = 'egrep "%s" "%s"' % (text, filename)
+        egrep_cmd = 'egrep "%s" %s' % (text, _expand_path(filename))
         return func(egrep_cmd, shell=shell).succeeded
```

This `_expand_path()` was then applied everywhere in contrib.files "just in case" but it was never needed anywhere, just quotes need to not be used around the filename, and they are already absent everywhere else.

As a sanity check, I tested with this fab task, and it worked as expected (both functions operated on `/home/pierce/test.txt`):

```python
@task
def test():
    files.append('~/test.txt', ['127.0.0.1 db', '127.0.0.1 redis'])
    files.sed('~pierce/test.txt', 'redis', 'mongo')
```

This is another one of those cases where unit tests don't really show correctness - good tests would need to actually run shell commands on the remote side of an ssh session, and then observe that after escaping and evaluating, the correct things happened to the correct files. Anyway, I'm pretty confident in this change - ` filepath ` should always be at least as good as ` "$(echo filepath)" ` for any conceivable purpose.


ploxiln:
I must admit, the integration tests caught me - these functions used to work with spaces, and with my change they don't.

But in my defense: they only worked with single spaces, and only if `~` was the only shell-expansion used. So `~/with space` would be expanded by the shell to `$HOME/with` and `space` and then "echo" would join the two arguments with a single space. If there were two spaces in a row they would be collapsed into one. If there were any spaces and a `*` or `?` were used, it would not work in a logically consistent or useful way - each space-separated part is expanded separately before being joined with a single space by "echo".

The simple way to get back to working at least as well as the "echo" trick is to backslash-escape the spaces in the filenames. As a bonus, the cases mentioned above (multiple space, or any space and other shell expansion) will work.

Finally, I must also admit that I forgot about the integration tests - these are a good way to test these contrib.files functions. (The unit tests for these functions, on the other hand, are not very useful: "the command strings generated match these samples that were generated before" rather than "these command strings actually work"). I definitely should add good integration tests for the stuff I've changed in here, sometime...

ploxiln:
... and if this is too much, I'd also be happy with just going back to #1614

PR #1616, opened 2017-06-01
Update installing.rst on paramiko dependency

veatch:
Clarify dependency differences between paramiko 1.x and 2.x.

### DIFF
--- a/sites/www/installing.rst
+++ b/sites/www/installing.rst
@@ -31,12 +31,13 @@ dependency)::
 Dependencies
 ============
 
-In order for Fabric's installation to succeed, you will need four primary pieces of software:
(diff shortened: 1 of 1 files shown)


bitprophet:
Thanks! Maybe we should just make this a link to Paramiko's install document, I'm surprised it isn't already one. (Especially since, with Cryptography in place, it's really dependencies plural, as Cryptography has a good half dozen or more sub-dependencies...)

veatch:
Yeah that makes sense - I just pushed another commit. Let me know if you think it's worth mentioning PyCrypto vs Cryptography or tweaking the wording.

bitprophet:
Perfect. Thanks!

PR #1631, opened 2017-07-14
keepalive maps to ServerAliveInterval

haridsv:
Currently, the documentation states that keepalive maps to ClientAliveInterval, which is a server side setting, unlike what the name indicates. The client side setting is called ServerAliveInterval. You can see references to this in the below two discussions:

https://github.com/paramiko/paramiko/issues/918
https://unix.stackexchange.com/a/3027/6475

### DIFF
--- a/sites/docs/usage/env.rst
+++ b/sites/docs/usage/env.rst
@@ -425,7 +425,7 @@ The global host list used when composing per-task host lists.
 **Default:** ``0`` (i.e. no keepalive)
 
 An integer specifying an SSH keepalive interval to use; basically maps to the
-SSH config option ``ClientAliveInterval``. Useful if you find connections are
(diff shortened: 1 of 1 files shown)


bitprophet:
Hrm this should've been fixed by #1065. Guessing git & moved paths strikes again? Thanks for noticing.

bitprophet:
Ah, apparently it was just a merge problem, it's there in 1.10 and 1.11 but not 1.12 upwards. Fixed that. Thanks again!

haridsv:
I see, you fixed it back in 2015 but it didn't get to master. Thanks for taking care of it!

bitprophet:
Technically I think the commit was 2015, but I only actually cherry-picked or merged it this past April. (Yea... 😭 ) Anyway, all good now regardless!

Issue #1591, opened 2017-04-21
Fabric 2 alpha/beta feedback meta-ticket!


bitprophet:
**Please make sure you've read this blog post and all its links first!:** http://bitprophet.org/blog/2017/04/17/fabric-2-alpha-beta

----

This is the place to leave feedback for issues not covered under any of the existing tickets; **please search the below places first!!**

- Fabric's own issues for anything strictly SSH related: https://github.com/fabric/fabric/issues
  - Especially the 'Invoke' label: https://github.com/fabric/fabric/labels/Invoke
  - And the 2.0 milestone: https://github.com/fabric/fabric/milestone/24
- Invoke's issues for things like tasks, subprocesses, CLI parsing, etc: https://github.com/pyinvoke/invoke/issues
- Lower level SSH things might be found in: https://github.com/paramiko/paramiko/issues

----

Not finding anything relevant? Leave a comment below! I'm looking for feedback _similar but not limited to_:

- "I refuse to use 2.0 until you put back in the magic, global, non object oriented mode!" (though I'll just link you to pyinvoke/invoke#186 and ask you to supply rationale there ;))
- "I really like what you did with $feature but it's missing $sub-feature, which I really need! Is that planned?" (I'll either say 'yes', 'no' or 'no but you can now trivially implement it yourself! it doesn't need to be in core!')
- "I really like what you did with $feature but the way it's set up right now, it's hard/impossible to do $sub-use-case!" (I'll probably ask you for details and possibly ask for an example PR of the API you'd like to see.)

UkuLoskit:
Small mistake in the URL, it should be http://bitprophet.org/blog/2017/04/17/fabric-2-alpha-beta/

bitprophet:
Goes to show how well I remember my own website, eh? Thanks! I even ran a link checker on my post, but...not against this ticket ;)

RedKrieg:
[...]

Issue #98, opened 2011-08-19
Optionally avoid using ssh if going to localhost


bitprophet:
### Description

run()/sudo() would intelligently see that you're going to localhost and just run local() instead.  This would probably be an optional thing.

Comments from Jeff on IRC:

<blockquote>

 and yea, I mean there's always going to be overhead with ssh vs straight pipes

 offhand I don't think it would be terrifically difficult to update run/sudo (especially in master now that they've been refactored) to call/return local() intelligently

 I'm not positive that I'd want that semi magical behavior in core (even with it off by default with an optin to enable it, tho that would help)

 but even so, it'd be an interesting experiment.

 and if it is as simple as I'm thinking I honestly can't come up with a good reason not to (again provided it is not the default behavior)
</blockquote>

---

Originally submitted by **Nick Welch** ([mackstann](https://github.com/mackstann)) on **2009-11-11** at **01:39pm EST**
### Relations
- Duplicated by #364: Allow for local operation to bypass SSH layer
- Related to #26: Implement "dry run" feature

bes89:
:+1:

laike9m:
+1

gerardohuck:
+1 Please implement this feature :)

cgarciaarano:
+1 

Could be very useful to bootstrap Docker images using existing Fabric scripts. This feature would avoid to install an SSH server on the container, which is against the [Docker best practices](https://docs.docker.com/articles/dockerfile_best-practices/)

F481:
+1

sbilly:
+1

cscetbon:
+1

foxx:
Further to the answer provided by @AntoniosHadji, here are the complete instructions to make this work;

```bash
# Generate new SSH key for local usage
ssh-keygen -f ~/.ssh/id_rsa -N ''

# Add server keys to users known hosts (eliminates 'are you sure' messages);
ssh-keyscan -H localhost > ~/.ssh/known_hosts

# Allow user to ssh to itself
cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
```

cgarciaarano:
Actually, this can be done using [cuisine](https://github.com/sebastien/cuisine). You need to change all ``run`` executions to reference ``cuisine.run`` function, which can be done easily with an import, and change the mode to local:

```python
from cuisine import run, mode_local

mode_local()
print run("echo Hello")
```

sbilly:
[...]

PR #1637, opened 2017-08-01
[wip] local vs remote task execution (ref #98)

rpkilby:
Hey @bitprophet. Some of these thoughts have been rattling around in my head since pycon, and I wanted to hash them out as thoroughly as possible before I forget. Mainly, I want to address the expectations of local/remote task execution [described](https://github.com/fabric/fabric/issues/1591#issuecomment-296915764) by @max-arnold and in #98, as well as the relationship of `Connection` and `Context` objects. Most of the thought work has already been done in #98, but I'm trying to distill that into a solution that is simple and non-surprising. In short,

1. There should be separate local/remote `task` definitions that always expect to operate on their respective `Context`/`Connection` objects.
2. `Connection`s should expose a separate `Context` that is explicitly used for local execution. 

----

**1:** Primarily, I want to argue that remote and local task execution should **not** be interchangeable, and by extension this should apply to `Connection` and `Context` objects. There should be separate local/remote `task` definitions that ensure that the correct context type is provided to the underlying task function.
- Interchanging Contexts/Connections with local/remote tasks seems more likely to be a user execution error than an expected outcome.
  - I would expect remote task execution to explicitly complain about the lack of a host, not to operate on the local environment and then *potentially* fail.
  - Similarly, I'd expect that local tasks should guard against remote execution. 
- More generally, fabric expects to operate on remote environments, and the context object needs to represent *execution* in remote environments. Invoke tasks should in turn expect a context that represents local execution. 
- That said, there are cases where execution is environment agnostic. Discussed in 1.1 below.

The differences in API (both current & future), kind of demonstrate that contexts and connections are not logically interchangeable.
- The `Connection.local` method is nonsensical in regards to `Context`s since they already operate on the local environment. As such, remote tasks that call `local` when operating on a `Context` will fail.
- `Connection`s will have `put`, `get`, and `reboot` operations that are similarly nonsensical for `Context`s.
- `Context`s can intermix regular python code with `run`/`sudo` calls. This behavior would be potentially incorrect on a remote `Connection`. eg, file writes would not occur on the remote server, OS info would not represent the remote server, etc...

I've started this PR as a wip demonstration for fabric, however this would also need corresponding changes to Invoke. What I'm thinking at the moment:
- Invoke should provide an abstract `BaseContext`, which all contexts inherit from.
- Invoke's `Context` should inherit `BaseContext` and more explicitly represent the local environment.
- Fabric's `Connection` should inherit `BaseContext` and more explicitly represent a remote environment.
- Alter `invoke.Task` to assert that it's first positional argument inherits `Context` (not `BaseContext`).
- Expose a `fabric.Task` that asserts its first positional argument inherits `Connection`.
- Expose a corresponding `fabric.task` decorator for `fabric.Task`s.
- Fabric should simply alias `invoke.task` as `fabric.local_task` (or something similarly appropriate).
- Should the fab executor differentiate between local and remote tasks? Maybe group them in the CLI?

**1.1:** As an extension of the above, one of the discussion points is the need to execute tasks in both local and remote environments. While this is certainly valid, how common and useful is this? Are there any non-trivial, more than just `ctx.run('whoami')` examples of this? I'd argue that this is more likely a minority of cases and it's not beneficial to cater towards it.
- Interchangeability/reusability should simply be handled with regular functions that expect any context.
- If a task truly is environment-agnostic, then it seems simple enough to either:
  - expose separate local/remote variants that wrap the same function.
  - expose a local task that can construct a Connection when appropriate.

**2:** `Connection`s are a `Context` for the sake of holding state information, but they should expose a separate `Context` for local environment execution. Change `Connection.local` to provide a *local* `Context` instead of simply aliasing the parent's `run` method. While technically backwards incompatible, the migration path should be straightforward. Altering @max-arnold's [example](https://github.com/fabric/fabric/issues/1591#issuecomment-296903582) might better illustrate this change.

```python
from fabric import task, local_task


@local_task
def build(ctx):
    with ctx.cd('/project/dir'):
        ctx.run('build > artifact.zip')


@task
def deploy(conn):
    build(conn.local)

    with conn.cd('/remote/path'), conn.local.cd('/project/dir'):
        conn.put(remote_path='build.zip', local_path='artifact.zip')
```

A few details:
- Given the API changes, both `fab -H host deploy` and `fab build` should now work as expected.
- This automatically fixes issues with `cd`/`prefix` interactions between the local and remote environment, since we're now operating on separate contexts/connections. No need for `cd` vs `lcd`.
- Easier invocation of local `sudo` commands.

**3:** Also a random, but related thought on executing `Connection('localhost').run()` as `local()`...
- Seems like a magical footgun. I'm expecting connections to operate over SSH. If I want to run local code, use `local()`. (or from the PR, `local.run()`)
- Additionally, the SSH config may contain user connection info. eg, ssh to localhost as a different user? idk.

----

Anyway, I'm definitely not married to the ideas in this PR as I've only spent a limited amount of time looking at a narrow subset of the problem. I'm sure I've missed some of the more varied/nuanced needs that have arisen over the years.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
   # TODO: real test matrix with at least some cells combining different invoke
   # and/or paramiko versions, released versions, etc
   # Invoke from master for parity
-  - "pip install -e git+https://github.com/pyinvoke/invoke#egg=invoke"
+  - "pip install -e git+https://github.com/rpkilby/invoke@context-runner#egg=invoke"
   # And invocations, ditto
   - "pip install -e git+https://github.com/pyinvoke/invocations#egg=invocations"
   # Paramiko ditto
--- a/fabric/__init__.py
+++ b/fabric/__init__.py
@@ -2,4 +2,5 @@
 from ._version import __version_info__, __version__
 from .connection import Config, Connection
 from .runners import Remote, Result
+from .tasks import local_task, task, LocalTask, Task
 from .group import Group, SerialGroup, ThreadingGroup, GroupResult
--- a/fabric/config.py
+++ b/fabric/config.py
@@ -5,7 +5,6 @@ import os
 from invoke.config import Config as InvokeConfig, merge_dicts
 from paramiko.config import SSHConfig
 
-from .runners import Remote
 from .util import get_local_user, debug
 
 
@@ -236,9 +235,6 @@ class Config(InvokeConfig):
             'port': 22,
             'run': {
                 'replace_env': True,
-            },
-            'runners': {
-                'remote': Remote,
             },
             'ssh_config_path': None,
             'tasks': {
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -101,6 +101,7 @@ class Connection(Context):
     connect_kwargs = None
     client = None
     transport = None
+    local = None
     _sftp = None
     _agent_handler = None
 
@@ -221,6 +222,10 @@ class Connection(Context):
         # that below. If it's somehow problematic we would want to break parent
         # __init__ up in a manner that is more cleanly overrideable.
         super(Connection, self).__init__(config=config)
+
+        #: A handle for a local Context that allows local commands to be ran
+        #: independently of the Connection's context.
+        self.local = Context(self.config.clone())
 
         #: The .Config object referenced when handling default values (for e.g.
         #: user or port, when not explicitly given) or deciding how to behave.
@@ -572,17 +577,6 @@ class Connection(Context):
         """
         runner = self.config.runners.remote(self)
         return self._sudo(runner, command, **kwargs)
-
-    def local(self, *args, **kwargs):
-        """
-        Execute a shell command on the local system.
-
-        This method is effectively a wrapper of `invoke.run`; see its docs for
-        details and call signature.
-        """
-        # Superclass run() uses runners.local, so we can literally just call it
-        # straight.
-        return super(Connection, self).run(*args, **kwargs)
 
     @opens
     def sftp(self):
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -410,6 +410,9 @@ class Connection(BaseContext):
         return hash(self._identity())
 
     def derive_shorthand(self, host_string):
+        if host_string is None:
+            return {'user': '', 'host': '', 'port': ''}
+
         user_hostport = host_string.rsplit('@', 1)
         hostport = user_hostport.pop()
         user = user_hostport[0] if user_hostport and user_hostport[0] else None
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -14,6 +14,7 @@ from paramiko.config import SSHConfig
 from paramiko.proxy import ProxyCommand
 
 from .config import Config
+from .runners import Remote
 from .transfer import Transfer
 from .tunnels import TunnelManager, Tunnel
 
@@ -222,6 +223,7 @@ class Connection(Context):
         # that below. If it's somehow problematic we would want to break parent
         # __init__ up in a manner that is more cleanly overrideable.
         super(Connection, self).__init__(config=config)
+        self._set(runner=Remote)
 
         #: A handle for a local Context that allows local commands to be ran
         #: independently of the Connection's context.
@@ -562,8 +564,7 @@ class Connection(Context):
             settings/behaviors; they are documented under
             `.Config.global_defaults`.
         """
-        runner = self.config.runners.remote(self)
-        return self._run(runner, command, **kwargs)
+        return super(Connection, self).run(command, **kwargs)
 
     @opens
     def sudo(self, command, **kwargs):
@@ -575,8 +576,7 @@ class Connection(Context):
         configuration overrides in addition to the generic/global ones. Thus,
         for example, per-host sudo passwords may be configured.
         """
-        runner = self.config.runners.remote(self)
-        return self._sudo(runner, command, **kwargs)
+        return super(Connection, self).sudo(command, **kwargs)
 
     @opens
     def sftp(self):
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -6,7 +6,7 @@ import socket
 from invoke.vendor.decorator import decorator
 from invoke.vendor.six import string_types
 
-from invoke import Context
+from invoke.context import BaseContext, Context
 from invoke.exceptions import ThreadException
 from paramiko.agent import AgentRequestHandler
 from paramiko.client import SSHClient, AutoAddPolicy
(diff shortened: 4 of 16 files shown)


max-arnold:
Just an idea:

```
from fabric import task, local

@task
@local
def build(ctx):
    with ctx.cd('/project/dir'):
        ctx.run('build > artifact.zip')

@task
def deploy(conn):
    build(local(conn))

    with conn.cd('/remote/path'), local(conn).cd('/project/dir'):
        conn.put(remote_path='build.zip', local_path='artifact.zip')
```

Basically `local()` can act both as decorator and function and transform `Connection` to `Context`. This seems to be more elegant than having separate `@task_local` decorator.

rpkilby:
Hi @max-arnold, thanks for the input. A few additional thoughts:
- For me, the important part is to separate the local context from the remote connection. I feel like this solves several related issues, and both of these proposals accomplish that.
- One consideration is that a `Connection.local` attribute would always provide the same `Context` object. A function/decorator/contextmanager would need return a fresh `Context`. I don't know if this difference matters or which of these would be desirable. 
- Is it possible to have an all-in-one function/context manager/decorator? `contextmanager` in python 3.2 (and backported to 2.6 with contextlib2) allows for combined decorator/context manager behavior, but I don't think it's possible for it to also operate as a regular function.

bitprophet:
Thanks for the writeup, Ryan!! I still need to kick this around in my head but here's my off the cuff thoughts after a first read.

---

I'm still on the fence about the overall core use case of #98:

- enough people have asked for it that it's clearly a need in one way or another, and while I haven't done it much myself, I totally grok the idea of having a deployment/config-management/etc task that one wants to test against a local system.
  - or for that matter, that one wants to invoke against a server from that server's own environment, e.g. how one runs `chef-solo` or `puppet` on a server itself.
- Requiring use of a local `sshd` still rubs me the wrong way, even as I find myself thinking "it's simply _not that hard_ to get one running no matter your platform".
  - Mostly because I know that even if you _can_ have an `sshd` running, there are legit reasons to want to skip it - the real world is messy and you may need/want to avoid complex authentication schemes, overly stringent allowed command sets, etc.
- So that leaves us with the primary challenge of how to enable writing code that can take arbitrary contexts without knowing if that context represents local or remote state (as you note yourself.)
  - You don't quite say this explicitly but I think it's implicit: the only real solution there is to encourage users to _write such functions/tasks against only the shared subset of local/remote API_ (which is to say, the methods which don't think in terms of two ends of a connection.)
  - In other words, `run`/`sudo`/`cd` and anything that builds on them, but not `put`/`get`/etc.

----

Your points about the API impedance mismatch (local contexts exposing nonsensical remote-only methods, or remote contexts growing a significantly different set of API members) are basically repeating things I've said before, so we're on the same page there (especially given what I just said above.) If we think in terms of "methods that care about one system/state, vs methods which care about two (or more)" that seems to make things clearer.

And I think I agree about the ideas surrounding the Context inheritance family & its use with the decorators. I'm not 100% sure we can't still use the current setup, with tweaks like _not_ renaming `run` to `local` but simply overriding it, but we'll see. This is mostly because at the moment, IIRC, most of the in-place remote-vs-local differentiation is in the Runner class hierarchy, with the Context->Connection subclassing being primarily additive. But I like the idea of being extra clean about it, so.

Also randomly wondering if we should go with `Context` -> `LocalContext` -> `SSHContext`. (This leaves space open for things like, idk, `TaskQueueContext` or perhaps hilariously, `TelnetContext` [those ancient embedded devices don't manage themselves!])

----

I think I like the general idea about Connection exposing a local Context as an attribute, especially re: the snippet showing how it enables explicit-but-easy, both-sides-of-the-connection state wrangling.

- I wonder if there's any other better method of obtaining a local Context object besides the Connection attribute, but I can't think of any, given the typical requirement to match the same config/CLI/etc state that the Connection would've been given by the execution machinery.
  - E.g. doing `local = Context(); local.blah()` loses all of that state and is thus far less useful. And it doesn't even save any typing or boilerplate over the attribute version of things.

----

> executing Connection('localhost').run() as local() [...] seems like a magical footgun

I mostly agree here, I think that idea largely came out as part of brainstorming. I'm much more likely to go for the "please write some single-context-expecting task or subroutine and hand it a `Context` of one stripe or another" option instead.

rpkilby:
So I finally decided to take a second stab at this. Note https://github.com/pyinvoke/invoke/pull/507, which implements the necessary underlying changes for this PR, and that this PR uses the corresponding branch for testing.

Overall, this PR is not complete but I think it's close. There are some issues with config creation in the executor, but I have run out of steam for the evening. It's probably best to kick this to you for feedback.

----

To recap, the fundamental idea of the PR is that tasks correspond to an execution environment. eg, an invoke task is tied to its `Context` and `Local` runner. Similarly, a fabric task should be tied to its `Connection` and `Remote` runner. With this, the `connection.local` is now a local context that can be passed to invoke tasks, or used to simply call `local.run` and `local.sudo` directly.

----

In general, it seems like we were on the same page. Addressing a few details:

> I'm still on the fence about the overall core use case of #98

One path forward would be to add a `LocalConnection` object that shares the same API as `Connection`, but which doesn't operate over an SSH connection. The executor could instantiate local connections when relevant, similarly to how it determines whens to create `Context`s and `Connection`s. This would be preferable to the `Connection` trying to perform some magic when operating on `localhost`. 

> Also randomly wondering if we should go with `Context` -> `LocalContext` -> `SSHContext`

I've opted for `BaseContext`, `Context`, and `Connection` in the PR, but naming aside, the relevant takeaway is that `SSHContext` inherits from the base `Context` instead of `LocalContext`. The reason is to help ensure that local tasks are not executed with an `SSHContext` instead of their expected `LocalContext`. ie, Invoke tasks expect invoke `Context`s, fabric tasks expect fabric `Connection`s. 

Hierarchy:

```
          Context                       BaseContext
         /       \           or         /         \
LocalContext    SSHContext        Context         Connection
````

bitprophet:
I have no recollection of closing this, wondering if the branch switch caused it automatically? I can't even try reopening it. @rpkilby if you rebasing against the new master and force pushing convinces this to reopen, let's do that, and otherwise I guess we gotta open a new one.

rpkilby:
I'm pretty sure this was automatically closed when the `v2` branch was closed. There was somewhere between 50 - 100 RPs automatically closed (quite a few notifications).

rpkilby:
I'll rebase in a bit.

bitprophet:
Great, thanks. FWIW I'm still not 100% sold on the approach here, but it's been on my mind and I haven't made a call one way or the other. We def need to solve it soon regardless. Possible I'll go the cheap route and do a smaller fix first and then go for this in 3.x. (Which will be way sooner than 2.x :P)

PR #1639, opened 2017-08-02
Update travis task names to be dasherized

rpkilby:
boop


rpkilby:
Ah, looks like there were some fairly significant changes to invoke beyond the dasherized task names.

bitprophet:
True, though this change should still be necessary, so thanks - might as well merge it :)

Issue #1633, opened 2017-07-18
fabric.api.get do not get the remote file mode


pysnow530:
When I download a executable file(0755) using `fabric.api.get()` from the remote server, the result file's mode is 0644. The executable flag is lost.

Maybe I lost an option? Or how could reserve the remote file mode?

bitprophet:
No, you're not wrong, `get` does not seem to have its own version of `put(mirror_local_mode=True)`. Thanks for the catch!

For what it's worth, you can of course use `os.chmod` to work around this in the meantime.

nikshepsvn:
@bitprophet is this still open? this would be my first open source contribution!

ploxiln:
looks like someone recently made an attempt at this in #1649

nikshepsvn:
is this still open? would love to give this a shot! can someone give me an idea of how to start off?

adrianyi:
it appears someone already implemented this, you can see how he did it [here](https://github.com/fabric/fabric/pull/1649/commits/bb08697452e801258f213412af777e023d94b756)

AdamDonna:
This appears implemented with https://github.com/fabric/fabric/pull/1649. Can we mark this as closed so new devs looking for "Low hanging fruit" don't get mislead?

pysnow530:
Thanks for reminding, @AdamDonna , closed!

pysnow530:
Thanks for reminding, @AdamDonna , closed.

PR #1649, opened 2017-08-22
mirror_remote_mode for api.get() file permissions

rkenmi:
In reference to Issue https://github.com/fabric/fabric/issues/1633

With get(..., mirror_remote_mode=True, ...), the files written locally will have the same file permissions as the remote file.

I would appreciate any kind of feedback/comments because I would like to implement this correctly!

Thanks!


bitprophet:
Looks good at a glance, thanks! Assigning to a milestone.

PR #1655, opened 2017-09-04
Implement a basic progress bar

gerases:
Hi, 

This is a very rough draft of a progress bar implementation. I just wanted to see what you think about this feature and whether you're interested. If you are, please let me know and i'll polish it off further based on your feedback.

One of the difficulties with the progress bar is that unless all output is funneled through the `write` method of `tqdm`, the bar will be broken. For that reason, I supplied a context manager for temporarily redirecting stdout to `tqdm`.

Obviously, a dependency on `tqdm` will be required.

Thanks,
  Sergei


gerases:
@bitprophet, it's hard for me to follow what happened with this pull request. Was it rejected or merged?

Issue #1661, opened 2017-09-07
local() does not set $SHELL


joshourisman:
Currently, if you wrap a command with the prefix context manager it executes using /bin/sh, but does not update the $SHELL environment variable accordingly. This becomes an issue if the context manager is wrapping a command that attempts to auto-detect the shell it's running in and alter behavior accordingly.

As a specific example, I have the following code in my fabfile:
```python
    with prefix('eval $(minikube docker-env)'), lcd(PROJECT_ROOT):
        docker_command = 'docker build -t {} {}'.format(local_image, service)
```

The `minikube docker-env` command attempts to detect your shell, and provides the necessary env variables for docker to use minikube (`eval` is obviously used to actually set those variables). Most of the time (for sh, bash, zsh, &c.), the shell detection is irrelevant as they all use the same `export` syntax to do this. However, fish does not, and if the fish shell is detected, the output will use `set -gx …` which simply doesn't work in sh (which prefix uses), causing the task to fail. As a result, if you're using fish and attempt to run that task, it won't work. (As it happens, all that's necessary is to change it to `minikube docker-env --shell=sh`, but there may be other circumstances in which there isn't such a simple option.)

To illustrate the issue, here's a contrived task that simply shows that even though the prefixed command is running in sh, the $SHELL env variable states that it's running in fish:

```python
@task
def test():
    with prefix('cd'):
        local('echo $SHELL')
```

The output of this task is:
```shell
$ fish
Welcome to fish, the friendly interactive shell
$ fab test                                                            
[localhost] local: echo $SHELL                                                                            
/usr/local/bin/fish                                                                                       

Done.   
```

ploxiln:
I think this is actually due to `local()` and how shells work, and not due to `prefix()`.

Try this: `bash -l -c 'echo $SHELL'`  - for you I expect it will output "/usr/local/bin/fish". A sub-shell only changes the SHELL variable if it is not set at all, i.e. if you do `unset SHELL ; bash -l -c 'echo $SHELL'` the output should be "/bin/bash".

`local()` does not directly choose the shell, by default it just lets `Popen` use its default "/bin/sh", but `local()` allows to override it. So this will probably work for you: `local(docker_command, shell="fish")`.

Notice that you will have this problem with anything that uses python standard `subprocess.Popen(... shell=True)`, it's not really specific to Fabric.

Maybe a reasonable fix is for `local()` to specify `env` to be a copy of the current environment vars, but with SHELL removed.

[...]

PR #1662, opened 2017-09-07
Have local() set the SHELL env var for consistent behavior.

joshourisman:
As show in #1661, the fact that `local()` (mistakenly stated as `prefix` in that issue) leaves the value of `$SHELL` unchanged can cause issues if the command being run depends on `$SHELL` referencing the actual shell being used. This pull request modifies `local()` to ensure that `$SHELL` is set appropriately. A test is included to ensure that `$SHELL` matches `$0` when the `local()` command is executed.


fabric/operations.py
```diff
@@ -1198,8 +1198,14 @@ def local(command, capture=False, shell=None):
...
     given_command = command
-    # Apply cd(), path() etc
+    # Apply cd(), path() etc, and set the SHELL env var to be consistent
+    set_shell = env.shell_env.get('SHELL')
+    env.shell_env['SHELL'] = shell or '/bin/sh'
     with_env = _prefix_env_vars(command, local=True)
+    if set_shell is None:
+        del env.shell_env['SHELL']
+    else:
+        env.shell_env['SHELL'] = set_shell
```
ploxiln: This is messy and unnecessary. Just copy the env, modify it, and pass it to `Popen()`:

```python
sub_env = os.environ.copy()
sub_env['SHELL'] = shell or '/bin/sh'
...
subprocess.Popen(cmd_arg, env=sub_env, shell=True, ...)
```

I still think just deleting SHELL from `sub_env` would be simpler, as it gets the actual shell to set the appropriate value, as bash does. I don't know if fabric really runs on windows (I think it theoretically does), but that would be an example of a case where `Popen()` would not actually use /bin/sh so assuming it does here would be wrong.

tests/test_context_managers.py
```diff
@@ -125,6 +125,19 @@ def test_cd_prefix():
         eq_(command_out, 'cd %s >/dev/null && foo' % some_path)
 
 
+#
+# prefix $SHELL
+#
+
+def test_prefix_shell_variable():
+    """
+    prefix should set the $SHELL varible for consistent behavior
```
ploxiln: this whole issue really has nothing at all to do with `prefix()`

joshourisman: Oops, you're right. Forgot to move that. :(

fabric/operations.py
joshourisman: Definitely unnecessarily messy, not sure what I was thinking. Will fix that.

I tried the approach of deleting SHELL from sub_env, but that doesn't actually seem to solve the problem. In my case, `local('echo $SHELL')` still returns `/usr/local/bin/fish`.

Windows not working with this is a good point.

ploxiln: ... might be because you tried with `env.shell_env` which is an additive-only fabric thing.

Anyway, thanks for the report and the PR, it's a good find :)

joshourisman:
Fixed the test, test_operations.py instead of test_context_managers.py, and doesn't use or reference prefix.

I also cleaned up the code actually setting the SHELL env var using the approach @ploxiln recommended. I tried passing in a clean env with no SHELL var set at all, but my local $SHELL value still came through.

joshourisman:
Oh, also just skipping modifications to env for win32. I don't currently have access to a Windows machine on which I could test this out, so this seems like the safest approach.

ploxiln:
Seems reasonable to me, I'd be fine with this.

But deleting SHELL from env works for me:

```
$ SHELL=fish python -c 'from subprocess import Popen; Popen("echo SHELL=$SHELL", shell=True)'
SHELL=fish

$ cat test.py 
import subprocess
import os

sub_env = os.environ.copy()
if 'SHELL' in sub_env:
    del sub_env['SHELL']

subprocess.Popen("echo SHELL=$SHELL", env=sub_env, shell=True)

$ SHELL=fish python test.py
SHELL=/bin/bash
```

joshourisman:
Interesting. If I run that exact same test.py, i get this:

```shell
$ python2 test.py 
SHELL=/usr/local/bin/fish
```

I'm running fish 2.6.0 and Python 2.7.13.

joshourisman:
Ah ha! It's apparently because you're explicitly setting SHELL when you run it. If I do that, I get the same result as you:

```python
$ SHELL=fish python2 test.py 
$ 
```

joshourisman:
Ok, nevermind. I have no idea what's going on.

```shell
$ bash
bash-3.2$ SHELL=fish python2 test.py
bash-3.2$ SHELL=/usr/local/bin/fish
```

ploxiln:
hmm could be `/etc/profile` or `~/.bashrc` or something ¯\\_(ツ)_/¯

PR #1664, opened 2017-09-14
Update for Py36 compatible changes

vineetvermait:


bitprophet:
I'm afraid master isn't yet Python 3 compatible, though further, I don't get why this is a 3.6 related change - a more-correct python 3 style change would be `except ssh.SSHException as err`, which we can't do yet as it'd break compat with Python 2.5 (old, yes, but still supported.)

2.0 (see the `v2` branch!) supports Python 3 and will be out sometime this year.

PR #1705, opened 2018-01-28
Add option to preserve timestamps on 'put' function

mjeveritt:
This is an updated version of @vitek 's Pull Request #544


mjeveritt:
I have a strong suspicion (untested) that one will run into the same issue with adding a parameter as discovered [here](https://github.com/fabric/fabric/pull/1706#issuecomment-361064045) in the 'get' function.

PR #1738, opened 2018-05-01
Add Connection.socks_proxy() and tunnels.SocksTunnelManager

steveprentice:
Opening this PR against the v2 branch because it seems like that is where active development is. Really love the changes you've made in v2!

This PR adds a socks5 compliant proxy tunnel manager and an associated contextmanager at Connection.socks_proxy().

Hoping to get your early feedback on this PR.

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -15,7 +15,7 @@ from paramiko.proxy import ProxyCommand
 
 from .config import Config
 from .transfer import Transfer
-from .tunnels import TunnelManager, Tunnel
+from .tunnels import TunnelManager, SocksTunnelManager, Tunnel
 
 
 @decorator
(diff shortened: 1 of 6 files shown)


steveprentice:
Failures:

> ValueError: Tried to add an argument named 'F' but one already exists!

I see this same error on the v2 branch, so I don't think it is related to my PR. However, if it is, let me know and I'll figure it out.

bitprophet:
Yea I literally just ran into the flag clash myself and pushed the fix, sorry about that! Try rebasing/merging.

bitprophet:
Thanks for this! I read some of the PR but am too mentally out of it right now to give the entire thing a thorough thumbs up/down review 😪  Stuff I did notice already:

- There's enough new code for the SocksTunnelManager that we might want to make `tunnels.py` into `tunnels/` the subpackage
- The `envvars` manager should live in Invoke, not Fabric, as it's local-context-specific
  - and it should probably be merged with [this similar-enough test-oriented fixture](https://github.com/pyinvoke/invoke/blob/4e95e8c9b2ffd16464aaf04bf378315c95a1e9dd/tests/conftest.py#L21-L26) from Invoke's test suite to boot
  - which is currently only about restoring any arbitrary env changes made within the managed block, but of course, applying changes can simply be an optional behavior
  - also of note, I _think_ that fixture's approach of "store old env, then wipe-and-restore after" is probably cleaner/safer than attempting to 'walk back' whatever changes were made inline? but lmk if I'm missing something there
- Is use of PySocks actually required? What happens if someone doesn't have it installed?
- Code style is a bit out of alignment with this repo's house style, I think it will even throw flake8 errors once the Invoke flag clash is resolved.

steveprentice:
Thanks @bitprophet. I'll push up some changes based on your initial feedback. A couple quick responses below:

> also of note, I think that fixture's approach of "store old env, then wipe-and-restore after" is probably cleaner/safer than attempting to 'walk back' whatever changes were made inline? but lmk if I'm missing something there

The behavior was intentional to avoid unexpected surprises. I originally had it the same as your test fixture. Consider the example below:

```python
with Connection('server').socks_proxy():
  region = requests.get('http://consul:8500/v1/kv/config/region').json()
  os.environ['AWS_DEFAULT_REGION'] = region

# Surprised os.environ['AWS_DEFAULT_REGION'] is set to the previous value or blown away.
```

One option would be to support both: if no specific envvars are passed in, then restore the entire environment. If specific vars are passed in, then restore just those.

> Is use of PySocks actually required? What happens if someone doesn't have it installed?

PySocks does not become a requirement of fabric2 with this change. However, `requests` does not support socks proxies without installing an additional library, as noted [in requests' documentation about using socks proxies](http://docs.python-requests.org/en/master/user/advanced/#socks): "In addition to basic HTTP proxies, Requests also supports proxies using the SOCKS protocol. This is an optional feature that requires that additional third-party libraries be installed before use."

> Code style is a bit out of alignment with this repo's house style, I think it will even throw flake8 errors once the Invoke flag clash is resolved.

`flake8` doesn't report any warnings or errors. If you can point out one or two instances of inconsistency, that'd get me started on know what you are expecting. I am trying to match the repo's style. Perhaps [paramiko/sftp.py](https://github.com/paramiko/paramiko/blob/7c7568605ed02c36995f9e51e0bc5c6dfdac909b/paramiko/sftp.py) is an example of what you are looking for? It has some similarities to this code. Packing, unpacking, protocol defined constants, etc.

steveprentice:
Congrats on the release! Should I reopen this against master or did you close because you're not interested in the PR?

bitprophet:
I wish github would mark automatic closes due to branch renames more clearly! I didn't close this myself and the reopen button is disabled. Think you have to file a copy of the PR against the new master branch - sorry!

To be clear: I'm still interested in the feature - tho might take a while to get to it, lots of fires to put out right now 😩

PR #1745, opened 2018-05-12
Fallback to system modules if vendorized one do not exist

athmane:
PS. this will help with fabric2 packaging.

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -1,10 +1,15 @@
 from contextlib import contextmanager
 from threading import Event
-from invoke.vendor.six import StringIO
+try:
+    from invoke.vendor.six import StringIO
+    from invoke.vendor.decorator import decorator
+    from invoke.vendor.six import string_types
+except ImportError:
+    from six import StringIO
+    from decorator import decorator
+    from six import string_types
 import socket
 
-from invoke.vendor.decorator import decorator
-from invoke.vendor.six import string_types
 
 from invoke import Context
 from invoke.exceptions import ThreadException
--- a/fabric/group.py
+++ b/fabric/group.py
@@ -1,4 +1,7 @@
-from invoke.vendor.six.moves.queue import Queue
+try:
+    from invoke.vendor.six.moves.queue import Queue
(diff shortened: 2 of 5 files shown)


bitprophet:
Thanks! I believe we do this in Invoke but I didn't get around to applying it to Fabric.

bitprophet:
Hm. Master or 2.0...master or 2.0...feels like arguably a bug so gonna backport.

bitprophet:
Also I'd love to test this but it seems more trouble than it's worth; any test would have to expect one setup or the other (existence of standalone `six`, or lack, for example) or be parameterized at runtime. And we've got bigger fish to fry right now.

rpkilby:
Out of curiosity, what's the reason to do this? In what cases would the vendored package imports not be present?

bitprophet:
@rpkilby There was a ticket for this for Invoke at one point, the tl;dr is that some OS level packages don't agree with vendoring and prefer to distribute a copy of the code that expects a "regularly" installed copy of those deps.

So e.g. Debian might yank out `vendor/` and add metadata depending on say `python-six` (which shows up in the Python import system as just regular ol `six`) and at that point we'd break without this change.

"Normal" installs from pip will always end up using the vendored copy, but then the code can be used by these downstream packages as-is without them having to apply patches.

*Note* that this is subtly different from the inverse, preferring the public copy of a module and only falling back to the vendored one, which has its own issues and is not desirable.

rpkilby:
> So e.g. Debian might yank out `vendor/` and add metadata depending on say `python-six` (which shows up in the Python import system as just regular ol `six`) and at that point we'd break without this change.

Interesting, I didn't realize that this occurred. Thanks for the info.

bitprophet:
If you're curious for more, go git blame in Invoke and you'll presumably end up at the old ticket about this, which might have more info. But that's my recollection.

athmane:
@rpkilby reasoning behind this is in:

https://github.com/pyinvoke/invoke/issues/204
https://github.com/pyinvoke/invoke/pull/412

PR #1755, opened 2018-05-14
[wip] Separating local and remote execution contexts

rpkilby:
This is a rebase of #1637 relative to the updated master branch. There is a lot of discussion/context in the original PR, but the gist is...

Invoke and Fabric tasks operate on fundamentally different execution environments, and their contexts should not be interchangeable. While you *may* be able to get away with running an Invoke task with a Fabric `Connection`, the inverse is not true - an Invoke `Context` cannot be promoted into a `Connection`. 

The primary change is how the `Connection` extends the `Context`, moving from an inheritance relationship to composition. Instead of the `Connection` being interchangeable, it should contain  a `Context` that is suitable for use with Invoke tasks.

One benefit is this simplifies the API for managing local vs remote actions. eg, because of the composed `.local` context, it would not be necessary to have both the `cd`/`lcd`, `run`/`local`, and other pairs of remote/local variants. Instead, the `Connection` and `Context` would have their separately defined APIs. eg, 

```python
with conn.cd('...'):
    with conn.local.cd('...'):
        conn.local.run('...')
    conn.run('...')
```

This PR also introduces a new Fabric `task` decorator (with Invoke's `task` imported as `local_task`), which expect their associated context type. e.g., it would not be possible to accidentally provide an Invoke context to a Fabric task, and a Fabric connection to an Invoke task. Note that these changes are dependent on pyinvoke/invoke#507.

The proposed usage would look something like:

```python
from fabric import task, local_task


@local_task
def build(ctx):
    with ctx.cd('/project/dir'):
        ctx.run('build > artifact.zip')


@task
def deploy(conn):
    build(conn.local)

    with conn.cd('/remote/path'), conn.local.cd('/project/dir'):
        conn.put(remote_path='build.zip', local_path='artifact.zip')
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
   # TODO: real test matrix with at least some cells combining different invoke
   # and/or paramiko versions, released versions, etc
   # Invoke from master for parity
-  - "pip install -e git+https://github.com/pyinvoke/invoke#egg=invoke"
+  - "pip install -e git+https://github.com/rpkilby/invoke@context-runner#egg=invoke"
   # And invocations, ditto
   - "pip install -e git+https://github.com/pyinvoke/invocations#egg=invocations"
   # Paramiko ditto
--- a/fabric/__init__.py
+++ b/fabric/__init__.py
@@ -2,4 +2,5 @@
 from ._version import __version_info__, __version__
 from .connection import Config, Connection
 from .runners import Remote, Result
+from .tasks import local_task, task, LocalTask, Task
 from .group import Group, SerialGroup, ThreadingGroup, GroupResult
--- a/fabric/config.py
+++ b/fabric/config.py
@@ -5,7 +5,6 @@ import os
 from invoke.config import Config as InvokeConfig, merge_dicts
 from paramiko.config import SSHConfig
 
-from .runners import Remote
 from .util import get_local_user, debug
 
 
@@ -244,9 +243,6 @@ class Config(InvokeConfig):
             'port': 22,
             'run': {
                 'replace_env': True,
-            },
-            'runners': {
-                'remote': Remote,
             },
             'ssh_config_path': None,
             'tasks': {
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -103,6 +103,7 @@ class Connection(Context):
     connect_kwargs = None
     client = None
     transport = None
+    local = None
     _sftp = None
     _agent_handler = None
 
@@ -223,6 +224,10 @@ class Connection(Context):
         # that below. If it's somehow problematic we would want to break parent
         # __init__ up in a manner that is more cleanly overrideable.
         super(Connection, self).__init__(config=config)
+
+        #: A handle for a local Context that allows local commands to be ran
+        #: independently of the Connection's context.
+        self.local = Context(self.config.clone())
 
         #: The .Config object referenced when handling default values (for e.g.
         #: user or port, when not explicitly given) or deciding how to behave.
@@ -586,19 +591,6 @@ class Connection(Context):
         """
         runner = self.config.runners.remote(self)
         return self._sudo(runner, command, **kwargs)
-
-    def local(self, *args, **kwargs):
-        """
-        Execute a shell command on the local system.
-
-        This method is effectively a wrapper of `invoke.run`; see its docs for
-        details and call signature.
-
-        .. versionadded:: 2.0
-        """
-        # Superclass run() uses runners.local, so we can literally just call it
-        # straight.
-        return super(Connection, self).run(*args, **kwargs)
 
     @opens
     def sftp(self):
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -14,6 +14,7 @@ from paramiko.config import SSHConfig
 from paramiko.proxy import ProxyCommand
 
 from .config import Config
+from .runners import Remote
 from .transfer import Transfer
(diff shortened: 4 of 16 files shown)


DylanYoung:
Why is the `local_task` decorator required?

rpkilby:
It's not really - more just a useful alias for fabric users to help distinguish between a local Invoke `task` and a remote Fabric `task`. Users would either need to do this:
```python
from fabric import task, local_task
```
or this:
```python
from fabric import task
from invoke import task as local_task
```

rpkilby:
It's been a while since I've thought about this PR. @bitprophet I'm happy to update/rebase/refactor etc if you're still interested in the direction this PR takes. I think it's an improvement, but obviously I'm a bit biased 😄.

DylanYoung:
What I mean is that there isn't really a need for the second import at
all.  There's already a way of going this: at the start of the task set
`cxt=ctx.local`.  It would also be nice to see a decorator shortcut:
`@task(delegate_to='local')`.  But the dual import seems unnecessary bloat.

DylanYoung:
I'd also like to see a `task(split_args=True)` to have the remote and local passed in separately, but that's just a nicety.

rpkilby:
> What I mean is that there isn't really a need for the second import at all. 

I'm not sure what you mean. With this PR, there would be two task types, an invoke/local task, and a fabric/remote task. You would need two imports for the two task types.

> There's already a way of doing this: at the start of the task set `cxt=ctx.local`

While `cxt=ctx.local` does work, the user just ends up with an `AttributeError` instead of an error that might be more explanatory. In contrast, this PR explicitly checks the task's context type. A fabric `task` **must** have a `Connection`, and an invoke `task` **must** have a `Context`. The reasoning is that local and remote tasks are intended to run in fundamentally different environments, and it is most likely a mistake if the user attempts to run a remote task on their local machine and vice versa. 

Note that this PR is dependent on [a corresponding invoke PR](https://github.com/pyinvoke/invoke/pull/507/files#diff-a2c23243533cad12946af2f1122bb4ebR109-R112), which would raise an error when the task is not provided the expected context type. Also, it's not part of this PR, but the difference in task types also means the executor and CLI could do more intelligent things with the various types of tasks (e.g., maybe the CLI groups tasks by type?)

> If you later find out that what *was* a local task now requires remote access,  then you need to change the whole way the task is written. With the delegate approach,  you just add some `cxt.remote...` calls and you're done.

I disagree. The API for the underlying `Context` doesn't change even when you promote the task type from invoke to fabric. If you start with:

```python
@invoke.task
def some_task(ctx):
    ctx.run(...)
```

And you want to then add some remote calls, it's as simple as:

```python
@fabric.task
def some_task(conn):
    ctx = conn.local
    ctx.run(...)

    conn.run(...)
```

DylanYoung:
What you just said makes little sense (first you said something was
impossible: setting cxt = cxt.local, then at the end you said it was
possible).  As such I think you're misunderstanding. See my comment on the
original issue for clarification. All this currently serves to do is
fragment the API.

For example: in a "local" task, you would do cxt.run. If you later require
remote work in that task,  all those calls must be rewritten to
cxt.local.run. In my proposed variant, you simply switch the default
delegate of the multi context (which contains both a remote and a local).

Side benefit: it enables future extensions to target multiple remotes with
a single task, without hacky and error prone switching.

DylanYoung:
In this scenario,  cxt is always an invoke cxt or extension of it.  If
invoke wants to add task "typing" or some such,  I think that's a side
issue, and certainly shouldn't be implemented as a different decorator per
type (imo), but rather as arguments to the task decorator itself.

rpkilby:
> What you just said makes little sense (first you said something was impossible: setting cxt = cxt.local, then at the end you said it was possible). As such I think you're misunderstanding. 

You're right, it sounds like we're not on the same page. Have you read the PR description in full?  The proposal boils down to two changes:

- Changing `Connection` to compose the local invoke `Context` instead of inheriting/extending.
- Differentiating remote from local tasks via distinct task decorators.

The rationale for these changes is given in the PR description, and a much more extensive discussion exists in the original PR. Given the above, I don't really understand this comment:

> What I mean is that there isn't really a need for the second import at all.  There's already a way of doing this: at the start of the task set `cxt=ctx.local`. 

I don't see how the logic follows. A `Connection` having access to the `local` context does not somehow negate the benefits of the two task decorators (which enforce that the task is executed against the proper environment).

---- 

> All this currently serves to do is fragment the API.

Yes, intentionally? Remote `Connection`s and local `Context`s operate on fundamentally different environments, and I argue that their APIs are not interchangeable. It's natural that one would support some operations that the other would not (e.g., a `Connection.reboot()` call). 

----

> For example: in a "local" task, you would do cxt.run. If you later require  remote work in that task,  all those calls must be rewritten to cxt.local.run. 

First, I don't think refactoring would be a big deal. Inside the function, You just search/replace `ctx.run` with `conn.local.run`. Second, as I stated in my previous comment, this is incorrect. Quoting myself:

> If you start with:
>
> ```python
> @invoke.task
> def some_task(ctx):
>     ctx.run(...)
> ```
> 
> And you want to then add some remote calls, it's as simple as:
> 
> ```python
> @fabric.task
> def some_task(conn):
>     ctx = conn.local
>     ctx.run(...)
> 
>     conn.run(...)
> ```

Adapting a local task to a remote task under this PR would be trivial. You only need to update the decorator and signature, and add one line to grab the local `ctx`.

PR #1763, opened 2018-05-16
Fix #1753

### LINKED ISSUE #1753: `Connection_.init.forward_agent.defaults_to_False` fails when ssh-config enables agent forwarding for `*`
For example, ssh-config contains this:

```
Host *
    ForwardAgent=yes
```

This manifests as follows:

```
================================================================================================================= FAILURES ==================================================================================================================
_____________________________________________________________________________________________ Connection_.init.forward_agent.defaults_to_False ______________________________________________________________________________________________

self = <connection.Connection_.init.forward_agent object at 0x10c59c048>

    def defaults_to_False(self):
>       assert Connection('host').forward_agent is False
E       AssertionError: assert True is False
E        +  where True = <Connection host=host>.forward_agent
E        +    where <Connection host=host> = Connection('host')

tests/connection.py:161: AssertionError
============================================================================================== 1 failed, 229 passed, 6 skipped in 2.63 seconds ==============================================================================================
```

offbyone:
The tests have a bug when a developer's host has its own ssh config
that sets agent forwarding to the nondefault.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,9 @@ before_script:
   - inv travis.make-sshable
 script:
   # Fast syntax check failures for more rapid feedback to submitters
-  - inv blacken --check
+  # (Python 3.6 only; it doesn't run on anything less and there's no
+  # point running it multiple times either.)
+  - "[[ $TRAVIS_PYTHON_VERSION == '3.6' ]] && inv blacken --check"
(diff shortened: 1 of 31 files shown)


bitprophet:
I took a different approach re: fixing this, see #1753. Meh.

PR #1766, opened 2018-05-16
Rupatro/invoke fabric with m flag

RupeshPatro:
Added __main__.py to enable python -m fabric functionality.
Added test case that checks version of fabric
Updated changelog.rst


bitprophet:
Did some tidying up and merged - thanks!

PR #1795, opened 2018-06-13
Support GSS-API authentication when ssh_config specified

### LINKED ISSUE #1794: Support GSS-API authentication if ssh_config mentioned it.
Now, I found fabric does not pass GSS-API related options at all to Paramiko even if the options in ssh_config exists.

Although user can pass the related parameter using `connect_kwargs` now, I think it's better to pass minimal required parameters to avoid users' misunderstanding. (At least 'gss_auth: True')

Xeite:
Pass required parameter for GSS-API to Paramiko when ssh_config have
'GSSAPIAuthentication yes' option.

Fix #1794

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -369,6 +369,10 @@ class Connection(Context):
                 self.ssh_config["identityfile"]
             )
 
+        gss_auth = self.ssh_config.get('gssapiauthentication', None)
+        if gss_auth and gss_auth == 'yes':
+            connect_kwargs['gss_auth'] = True
+
         return connect_kwargs
 
     def __repr__(self):


bitprophet:
Thanks for this!

I don't have easy ability to test GSSAPI/Kerberos on my end, so I have to ask this on faith: is this all that's required or are there other related settings we want to be inspecting too? For example, I note the existence of `GSSAPIDelegateCredentials`, as well as things like possibly having `gssapi-with-mic` in `PreferredAuthentications`. (That latter may want to wait for paramiko/paramiko#387 tho.)

bitprophet:
Also, the changelog and the docs need an update for this, as do the tests. E.g. update http://docs.fabfile.org/en/2.1/concepts/configuration.html#connection-s-use-of-ssh-config-values as well as the relevant tests that are probably in either `tests/config.py` or `tests/connection.py`.

I may do this at merge time but if you want to take a stab at it please do.

Xeite:
Hi, thanks for kind guide.

I'm not fully sure that I understand your point which means the question is about integration with paramiko API or general kerberos requirements, though. Maybe I think you are already familiar with paramiko, so I try to answer the latter one.

Yes, there are few more options we have to think about. I found few options related to GSSAPI from https://linux.die.net/man/5/ssh_config. 

> 
> GSSAPIAuthentication
> Specifies whether user authentication based on GSSAPI is allowed. The default is ''no''. Note that this option applies to protocol version 2 only.
> 
> GSSAPIKeyExchange
> Specifies whether key exchange based on GSSAPI may be used. When using GSSAPI key exchange the server need not have a host key. The default is ''no''. Note that this option applies to protocol version 2 only.
> 
> GSSAPIClientIdentity
> If set, specifies the GSSAPI client identity that ssh should use when connecting to the server. The default is unset, which means that the default identity will be used.
> 
> GSSAPIDelegateCredentials
> Forward (delegate) credentials to the server. The default is ''no''. Note that this option applies to protocol version 2 connections using GSSAPI.
> 
> GSSAPIRenewalForcesRekey
> If set to ''yes'' then renewal of the client's GSSAPI credentials will force the rekeying of the ssh connection. With a compatible server, this can delegate the renewed credentials to a session on the server. The default is ''no''.
> 
> GSSAPITrustDns
> Set to ''yes to indicate that the DNS is trusted to securely canonicalize'' the name of the host being connected to. If ''no, the hostname entered on the'' command line will be passed untouched to the GSSAPI library. The default is ''no''. This option only applies to protocol version 2 connections using GSSAPI.
> 

The only thing that I added and tested was `GSSAPIAuthentication` option and other options set by default. Although I wanted to try to implement other options, our system was not allowed to any other options except `GSSAPIAuthentication`, so I could not test...T_T. Sorry for that.

For the testcase, I also could not find the way to make mocking `.ssh/config` and it seems to use system default one, so I am not sure how to correctly make a test. Is there any guide to make testcase? If then, I will happily add them.

Thanks.

yarikoptic:
FWIW, I have tested it and it seems to work perfectly for me  (having paramiko 2.7.1 and using this PR with a merged state (4ad6e60fae8054dd7aed07e11842b1b806a00d80 locally) of current master 2.5.0-8-g6824b103)!  Any chance to see this PR going forward? ;)

<details>
<summary>my ~/.ssh/config</summary> 

```shell
$> cat ~/.ssh/config
Host discovery7 discovery*.hpcc.dartmouth.edu
  GSSAPIAuthentication yes
  GSSAPIDelegateCredentials yes
```
</details>

yarikoptic:
ping on this PR?

PR #1825, opened 2018-07-14
Update link in README

jscissr:
http://docs.fabfile.org/en/latest/upgrading.html -> http://www.fabfile.org/upgrading.html#upgrading


bitprophet:
Good catch, thanks!

PR #1827, opened 2018-07-17
Update README.rst

hjb912:


bitprophet:
Thanks for the input! But why the change? The intent is that most principal API members can be imported directly from the package root (they're all explicitly imported to `__init__.py`) - it's simply less typing, less remembering the submodule in play, and easier to read. Going to close on those merits but always open to counterarguments :)

hjb912:
This is the case that my friend use some books that teaching 1.x use `from fabric.api import *`, however he use pip to install fabric 2.x,  `not find api`, so he watch the doc but failure to run it , and give up, because he said there is no `from fabric import Connection`, and we find out that it is`from fabric.connection import Connection`, so i think this is a good tools and make this request...= =~

Issue #1516, opened 2016-10-06
Consider enabling SSL for website(s)


dedeibel:
Just reporting, not judging, _flies away_

> www.fabfile.org uses an invalid security certificate. The certificate is only valid for the following names: *.readthedocs.org, readthedocs.org 

![fab-https-2016-10-06](https://cloud.githubusercontent.com/assets/26708/19144775/c4e9eede-8bab-11e6-998e-ae48b4c3ff76.jpg)

bitprophet:
Thanks for the note!

Yea, I don't think I ever bothered enabling SSL on ReadTheDocs; the non-HTTP domains don't auto-forward or anything, and I personally never share HTTPS-style links. I.e. - the HTTP URLs are "canonical" at the moment.

Looks like the right way to do it is to host the SSL myself and proxy to RTD: [link](http://docs.readthedocs.io/en/latest/alternate_domains.html#cname-ssl). This adds more of a support burden on my personal VPSen:

- right now, if my VPS goes down, IIRC the only thing affected is the apex domain, and explicit accesses to `docs.whatever.org` or `www.whatever.org` go directly to RTD and still work.
- and of course, the actual SSL certificate generation/setup/hosting/renewal/etc/etc (though I'd probably take a stab at using Let's Encrypt!).
- if I do this for Fabric I'll also want to do it for my other projects :) (not that this is a strong argument against, but still.)

dedeibel:
I am using ssl everywhere plugin which made me stumble upon it.

ploxiln:
If you're willing to use a subdomain of readthedocs.io for the documentation pages, https://fabric-docs.readthedocs.io/en/latest/ also works

bitprophet:
My DNS host (though not, for most domains, my registrar - I may change that sometime) just opened a beta of Let's Encrypt integration: https://blog.dnsimple.com/2016/11/lets-encrypt/

Haven't dug into how much easier it would/would not make this, assuming it'd remove some of the burden of LE automation on my server's end? ¯\_(ツ)_/¯

ploxiln:
just throwing out another option: github pages recently added https for custom domains

DilumAluthge:
@bitprophet Any update on this? Happy to help out.

bitprophet:
It's at the top of my personal priority list, given Chrome's recent updates ;) Hoping to update my VPS (or convert to GHP or whatever) soon!

bitprophet:
[...]

PR #1856, opened 2018-08-17
fix https issue with links to invoke and paramiko

satyapramodh:
As those domains are currently not on https, they result in 404 pages.


bitprophet:
See #1516, I'd like to just get those apex domains on HTTPS soon. Thanks though!

Alternately we could put in explicit `www` subdomains (which now do work on HTTPS after recent RTD updates). But I'd rather just get LE on my VPS. (So many acronyms...)

PR #1894, opened 2018-10-23
Dependent tasks are now executed per host.

StuartMacKay:
The run() and sudo() methods on a Connection give the expectation that the command will be run on the target host(s). However if a task is specified as a pre or post dependency then the command is run on the machine where the fab command was run.

This PR modifies the calls in the Executor.expand_calls() method to pass the value of allow_hosts to any dependent tasks. The result is that instead of the dependent tasks being executed only once on source machine they are executed on each target host.

### DIFF
--- a/fabric/executor.py
+++ b/fabric/executor.py
@@ -23,8 +23,7 @@ class FabExecutor(Executor):
             # handle intersect of dependencies+parameterization, just becomes
             # 'honor that new feature of Invoke')
             # TODO: roles, other non-runtime host parameterizations, etc
-            # Pre-tasks get added only once, not once per host.
-            ret.extend(self.expand_calls(call.pre, apply_hosts=False))
+            ret.extend(self.expand_calls(call.pre, apply_hosts=apply_hosts))
             # Main task, per host
             for host in hosts:
                 ret.append(self.parameterize(call, host))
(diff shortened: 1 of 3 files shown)


micahjsmith:
@StuartMacKay thanks for this fix, the referenced issues are still open as of fabric v3. would you want to change the base branch to `main` and see whether your patch still applies?

I'm not a maintainer, just hoping that this can still be merged.

davidjmemmett:
Hi, there is no fabric v3, that (fabric3) was a fork which is no longer maintained. Please use the latest fabric v2.Cheers,DavidOn 7 Apr 2023, at 21:35, Micah Smith ***@***.***> wrote:﻿
@StuartMacKay thanks for this fix, the referenced issues are still open as of fabric v3. would you want to change the base branch to main and see whether your patch still applies?
I'm not a maintainer, just hoping that this can still be merged.

micahjsmith:
@davidjmemmett sorry I'm a bit confused, the latest release of `fabric` is `v3.0.0`. the version on `main` branch is `3.0.0`. are you suggesting these are no longer maintained?

davidjmemmett:
You're correct - @bitprophet sneaked that out of the door!

micahjsmith:
@davidjmemmett okay, interesting, thanks -- created #2258 to continue discussion to avoid hijacking this thread

in that case, my comment about rebasing on main/v3 no longer applies, and I'd love to see this fix reviewed/merged. let me know if there is anything I can do to push this along, though I'm a new contributor to fabric

PR #1911, opened 2018-11-30
Update URLs for Invoke and Paramiko websites

andreroggeri:
Without the www the websites throws ERR_CONNECTION_REFUSED


anarcat:
#1960 is a more complete fix for this problem, and targeting the right branch (master). this pr should be closed.

PR #1927, opened 2019-01-14
Support put()/get() with sudo.

rectalogic:
Requires passwordless sudo access on remote.
Also you must know the path to the sftp subsystem on the remote.

Similar to:

`sftp -s "sudo /usr/lib/openssh/sftp-server" user@machine`

```python
from StringIO import StringIO
from fabric import task


@task
def putsudo(c):
    c.put(StringIO("fab put sudo"), "/root/fabsudo.txt", subsystem="/usr/bin/sudo /usr/lib/openssh/sftp-server")
```


lilcheti:
this is so needed :')

PR #1931, opened 2019-01-16
Fixed "Socket is closed" error when transferring files with new connection

thomasantony:
`_sftp` being a class variable does not get reset if I create a new Connection after closing an existing connection. The memoisation causes the old `_sftp` to be used which results in an error when transferring files on the new connection. 

This change ensures that the variable is cleared when the connection is closed.

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -686,6 +686,7 @@ class Connection(Context):
             self.client.close()
             if self.forward_agent and self._agent_handler is not None:
                 self._agent_handler.close()
+        self._sftp = None
 
     def __enter__(self):
         return self


graingert:
related to https://github.com/fabric/fabric/issues/1981

fabric/connection.py
```diff
@@ -686,6 +686,7 @@ def close(self):
             self.client.close()
             if self.forward_agent and self._agent_handler is not None:
                 self._agent_handler.close()
+        self._sftp = None
```
graingert, line 689: should this also close the _sftp client?

```python
if self.is_connected:
    sftp, self._sftp = self._sftp, None
    if sftp is not None:
        sftp.close()
    self.client.close()
```

davidjmemmett:
Thank you for this PR - I've also [submitted a PR](https://github.com/fabric/fabric/pull/2195) including a couple of tests and a quick documentation update.

PR #1950, opened 2019-02-28
Fix broken links and use https wherever possible

kbd:
I noticed some of the links on http://docs.fabfile.org/en/latest/getting-started.html were broken because of no 'www' on https links. Rather than fix things half-way, I went through the repo and changed everything possible to https and corrected links.

There are still a few regular http: urls left because those urls don't support https:

```
fabric[master]$ rg http:
sites/www/installing.rst
9:Fabric is best installed via `pip <http://pip-installer.org>`_::

sites/www/installing-1.x.rst
16:Fabric is best installed via `pip <http://pip-installer.org>`_; to ensure you

sites/www/development.rst
6:<http://bitprophet.org>`_, aka ``bitprophet``.  However, dozens of other
```

http://pip-installer.org forwards to https://pip.pypa.io/en/stable/ but I didn't want to change the link on you.

Btw, the repository defaults to the 2.0 branch, which was a little confusing. This PR is against master.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,11 +3,11 @@ Welcome to Fabric!
 
 Fabric is a high level Python (2.7, 3.4+) library designed to execute shell
 commands remotely over SSH, yielding useful Python objects in return. It builds
-on top of `Invoke <http://pyinvoke.org>`_ (subprocess command execution and
-command-line features) and `Paramiko <http://paramiko.org>`_ (SSH protocol
+on top of `Invoke <https://www.pyinvoke.org>`_ (subprocess command execution and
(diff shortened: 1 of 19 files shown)


kbd:
Looks like CI failed on Python 3.6 only because of a black "would reformat" check (even though the line is 80 chars long). So much for my "I don't need a branch for my pull request because it's only one commit". Lemme know if you want me to fix it...

anarcat:
> Looks like CI failed on Python 3.6 only because of a black "would reformat" check (even though the line is 80 chars long). So much for my "I don't need a branch for my pull request because it's only one commit". Lemme know if you want me to fix it...

please do fix it. so far, #1960 is better because it passes tests, but yours is more thorough.

PR #1960, opened 2019-04-04
Broken links in documentation

joshourisman:
Scattered throughout the documentation are a handful of broken links to both paramiko.org and pyinvoke.org. After some investigation it turns out that this is because both URLs https://paramiko.org and https://pyinvoke.org fail to connect. There is also an inconsistent usage of http and https throughout. I ran a simple find and replace to normalize all links to both domains to use https and the www subdomain to avoid any problems.


anarcat:
ping? could we get unit tests to pass here at least?

PR #1989, opened 2019-07-01
Add support for `command_timeout` in `run()`

fruch:
depends on:
https://github.com/pyinvoke/invoke/pull/645

### DIFF
--- a/fabric/runners.py
+++ b/fabric/runners.py
@@ -32,7 +32,7 @@ class Remote(Runner):
         self.inline_env = kwargs.pop("inline_env", None)
         super(Remote, self).__init__(*args, **kwargs)
 
-    def start(self, command, shell, env):
+    def start(self, command, shell, env, command_timeout=None):
         self.channel = self.context.create_session()
         if self.using_pty:
             rows, cols = pty_size()
@@ -55,8 +55,8 @@ class Remote(Runner):
                 command = "export {} && {}".format(parameters, command)
             else:
                 self.channel.update_environment(env)
-        # TODO: pass in timeout= here when invoke grows timeout functionality
-        # in Runner/Local.
+
+        self.channel.settimeout(command_timeout)
         self.channel.exec_command(command)
 
     def read_proc_stdout(self, num_bytes):


bitprophet:
My TODO:

- [ ] manual test
- [ ] unit tests
- [ ] maybe an integration test?
- [ ] changelog entry
- [ ] update Upgrading docs page to note that this is ported to v2 now
- [ ] This probably requires bumping Invoke dependency
- [ ] Also update changelog for previously merged stuff that mentions "works with invoke<1.3 but requires >=1.3 for xxx" since that'll be moot now...

bitprophet:
This may not be a simple merge either; the socket timeouts bubble up into the thread-exception handlers (as they occur on attempts to recv/recv_stderr in the IO workers).

Fabric 1 did something kinda similar but it had all the timeout logic meshed in the middle of the IO workers, so it was able to say "is this a socket timeout and are we past the configured timeout (to prevent muting connection-time timeouts, I think)? turn it into a timeout exception and raise that".

Looking at Invoke, I notice we do something vaguely similar with watcher errors - it definitely follows the obvious/naive option for us here, which is "examine thread errors to see if they need special handling". This might be a good excuse to cut up Runner some more so we can extend this pattern for Fabric.

----

The only other route I see offhand is to change the contract for timeouts and say "actually, the abstract Runner is what handles the timeout itself, via the threading Timer, and subclasses only need to ensure we have a way of forcibly stopping the subprocess from the main thread".

This way, we're not relying on "action at a distance" via the socket timeouts, but we just extend the Runner's main-thread wait loop - the same spot that handles KeyboardInterrupt (aka "the human got tired of waiting" style timeouts...:D) could look for timer expiry and call `self.send_interrupt`. It would also have to set some bookkeeping such that shutdown triggers the CommandTimedOut instead of just pseudo-gracefully stopping.

This is tempting partly because the whole "subclasses determine what timeout means" felt wrong to me - why would the execution mechanism change the fact that we want control returned to us after N seconds? I guess there are some situations, like accounting for transmission time of network exec (timeout defined as time spent running remotely only) or eg allowing a subclass to do something very different like wrapping the command in the `timeout` program, etc.

bitprophet:
Think I'll aim for the latter, started blocking out the former and it requires some mild contortions within Runner and its state (there's a reason _run_body is such a big single function right now...for better and mostly worse), I suspect the latter approach is going to be simpler to implement as well as "feeling more correct".

bitprophet:
`Channel.shutdown(2)` doesn't suffice, I believe because it doesn't actually close the connection & that means `Channel.exit_status_ready` blocks until the natural end of the process anyways. We may just want to `.close` instead - unhappy I don't know off the top of my head. EDIT: seems like that does the trick better.

bitprophet:
Almost forgot to check up on CLI args, glad I did - now invoke and fabric 2 retain the fabric 1 `-T` and `-t` flags. This is just about ready to merge.

bitprophet:
All done? Finally?

Issue #1886, opened 2018-10-03
Feature request: echo mode with server name


SwampFalc:
I am using invoke to drive a specific tool. Usually, I have it on pretty quiet but earlier today I needed a bit more information about what was happening, so I started using the -e flag to invoke.

Now, part of what's happening is that invoke loads fabric and makes a Connection to a remote server. I also wanted to see what was happening there.

It was very easy to basically pass that -e flag into fabric by passing the current invoke context to the Connection constructor.

However, this leaves me with echo output coming from fabric that is indistinguishable from the echo output of invoke... Far better than nothing, of course, but it could be better.

Would you consider adding something like `[server]` to echo output? Perhaps the echo parameter could be expanded to take either True, False or 'server'? Or allow it to take a template string, even?

exhuma:
I have the same issue and started working on a small patch for this. I will push it today and open a PR.

Additionally, I have the issue that for some reason the config value `run.echo` is ignored on the remote machines. It seems that the patch I made also solved this (but I am not 100% sure yet, which is why it is not yet pushed).

PR #2009, opened 2019-09-09
Prefix the "echo" output with the current hostname

exhuma:
References fabric/fabric#1886

Note that this depends on pyinvoke/invoke#662


exhuma:
Hmm... from what I can tell, the failing check is unrelated to these changes.

Issue #1992, opened 2019-07-10
Not thread safe, no async running options


abegus:
Either I'm missing something, or fabric2 / fabric is just a limited wrapper of paramiko? In paramiko you can run multiple things using one connection in parallel, and you can run a command asynchronously.

This blocks almost every use case I have for this package (which I was hoping was paramiko with better API's).

EX, if I have a proxy server that has a local connection to ~10 other servers, I must make 10 separate connections to that proxy server just to tunnel through to the other 10 servers (which itself will fail because of "paramiko Administratively prohibited", which is an issue caused by opening up too many connections to a server at one time.)

If I try to share a connection (which should be possible if fabric was thread safe), I would not get this issue on the second connection:

  File "<decorator-gen-3>", line 2, in run
  File "C:\ProgramData\hydra\hydra_workspace\autotest-venv\lib\site-packages\fabric2\connection.py", line 29, in opens
    self.open()
  File "C:\ProgramData\hydra\hydra_workspace\autotest-venv\lib\site-packages\fabric2\connection.py", line 615, in open
    self.client.connect(**kwargs)
  File "C:\ProgramData\hydra\hydra_workspace\autotest-venv\lib\site-packages\paramiko\client.py", line 437, in connect
    passphrase,
  File "C:\ProgramData\hydra\hydra_workspace\autotest-venv\lib\site-packages\paramiko\client.py", line 749, in _auth
    raise saved_exception
SSHException: No existing session

Issue #1889, opened 2018-10-09
`Fatal error: No existing session` when using execute()


gvfnix:
Hi, I have some strange issue when using Fabric 1.14 with python 2.7. If the code contains `execute()` statement, then fabric cannot open SSH connection to the host and the script fails with `Fatal error: No existing session`. But if I do not use `execute()`, the script runs normally.

Code sample:
```
#!/usr/bin/env python
#encoding: utf-8
import os, sys, logging
logging.basicConfig(level=logging.DEBUG)
from fabric.api import (execute, run, task, sudo, env, roles, hosts)

env.user="root"
env.password="***"
env.reject_unknown_hosts = False

@task
@hosts('10.100.0.100')
def getnproc():
    # cores = int(sudo('nproc'))
    # ram = int(sudo("cat /proc/meminfo | grep MemTotal | awk '{ print $2 }'")) / 1024
    # return {'cores': cores, 'RAM': ram}
    return run('nproc')

a = execute(getnproc, host='10.100.0.100')
```
Running the script:
```
[root@38768029c5f2 stress]# fab getnproc
[10.100.0.100] Executing task 'getnproc'
[10.100.0.100] run: nproc
DEBUG:paramiko.transport:starting thread (client mode): 0x8f161810L
DEBUG:paramiko.transport:Local version/idstring: SSH-2.0-paramiko_2.4.2

Fatal error: No existing session

Underlying exception:
    No existing session

Aborting.
DEBUG:paramiko.transport:Remote version/idstring: SSH-2.0-OpenSSH_7.4
INFO:paramiko.transport:Connected (version 2.0, client OpenSSH_7.4)
DEBUG:paramiko.transport:EOF in transport thread
```

But if I comment out the line `a = execute(getnproc, host='10.100.0.100')`, this script runs normally:

```
[root@38768029c5f2 stress]# fab getnproc
[10.100.0.100] Executing task 'getnproc'
[10.100.0.100] run: nproc
DEBUG:paramiko.transport:starting thread (client mode): 0xf27b27d0L
DEBUG:paramiko.transport:Local version/idstring: SSH-2.0-paramiko_2.4.2
DEBUG:paramiko.transport:Remote version/idstring: SSH-2.0-OpenSSH_7.4
INFO:paramiko.transport:Connected (version 2.0, client OpenSSH_7.4)
[...]

PR #2010, opened 2019-09-11
Make connection.open thread safe

crhan:
When connection shared between threads and theses threads start run a
command at the same time. The connection.open will be called multiple
times and they will all do the 'client.connect' and run into a strange state.

I think this problem is related to #1992 and #1889


breisig:
This would be good to get merged.

efagerberg:
Any update on this? I tested this out by changing the code in my virtualenv and it fixed the issue it seems. Would be good to get this in proper.

Issue #2007, opened 2019-08-31
AttributeError: 'str' object has no attribute 'extend' when using custom connect_kwargs and key_filename option


joeydumont:
When using a custom `connect_kwargs` with a `key_filename` option, like so:

```python
    config = fabric.config.Config()
    print("Connecting with ssh/config")
    with fabric.Connection(
        host=instancePrivateIp,
        user='ec2-user',
        config=config,
        connect_kwargs={
            "key_filename": keyFileName,
        }) as c_instance:
```
I get an `AttributeError`:

```python
  File "/home/dumontj/Projects/ami-builder/ami_builder/customization.py", line 26, in do_bash_customization
    "key_filename": keyFileName,
  File "/home/dumontj/.local/share/virtualenvs/ami-builder-jbdgQc96/lib/python3.7/site-packages/fabric/connection.py", line 450, in __init__
    self.connect_kwargs = self.resolve_connect_kwargs(connect_kwargs)
  File "/home/dumontj/.local/share/virtualenvs/ami-builder-jbdgQc96/lib/python3.7/site-packages/fabric/connection.py", line 490, in resolve_connect_kwargs
    connect_kwargs["key_filename"].extend(
AttributeError: 'str' object has no attribute 'extend'
```

The documentation specifies that `key_filename` should be given as a string, but I found that it works if you actually give it a list:

```python
    with fabric.Connection(
        host=instancePrivateIp,
        user='ec2-user',
        config=config,
        connect_kwargs={
            "key_filename": [keyFileName,],
        }) as c_instance:
```

If this is the wanted syntax, I can update the documentation through a PR (see https://github.com/fabric/fabric/blob/2.0/fabric/connection.py#L234). Otherwise I could change `fabric/connection.py` to check whether there is already an entry, then transform it into a list if necessary.

boweevil:
I'm seeing the same issue on 2.5.  However, using a list works.

adq:
Same problem. Looks like it could be a silly bug to be, so pinning to 2.4.0 till we hear more from upstream.

m4g005:
I get the same problem when host is set to a DNS name. When set to an IP, the exception doesn't occur.
E.g.:
```
{'host': '127.0.0.1', 'user': 'MyUser', 'connect_kwargs': {'key_filename': '/Users/MyUser/.ssh/MyPem.pem'}} # <- Works
[...]

PR #2014, opened 2019-09-18
Update docs to expect key_filename to be a list

RoryCrispin:


bitprophet:
Thanks, but this is no longer necessary after fixing #2007 at the source!

PR #2029, opened 2019-10-25
Fixes #1948 - Removes mentions of scp

### LINKED ISSUE #1948: Remove SCP references in Fabric doc
I was looking for a way to use SCP via Fabric, and after some investigation found issue #945 (Allow option for get/put to be handled by scp), currently closed.
> See [Paramiko #150](https://github.com/paramiko/paramiko/issues/150) - Fabric won't be able to take any action on this until we decide how/whether Paramiko should.
> 
> _Originally posted by @bitprophet in https://github.com/fabric/fabric/issues/945#issuecomment-26168826_

That was in 2013, and the question still seems open regarding how to do SCP instead of SFTP.
https://github.com/fabric/fabric/blob/e9939d68b734935f0c98d98817912ad7c698238f/fabric/transfer.py#L26

Given that Fabric cannot use SCP for transfers, can the documentation be updated to reflect that by omitting SCP? Might save others some investigative effort.

Thanks for your time!

PS: I'm still using Fabric for SSH workflows. Very easy to use! 👍

karlcow:
The documentation mentions scp in a couple of places, but scp is not working. This removes the text to avoid confusion for users.

### DIFF
--- a/fabric/transfer.py
+++ b/fabric/transfer.py
@@ -1,5 +1,5 @@
 """
-File transfer via SFTP and/or SCP.
+File transfer via SFTP.
 """
 
 import os
@@ -64,7 +64,7 @@ class Transfer(object):
 
             **If a string is given**, it should be a path to a local directory
             or file and is subject to similar behavior as that seen by common
-            Unix utilities or OpenSSH's ``sftp`` or ``scp`` tools.
+            Unix utilities or OpenSSH's ``sftp`` tool.
 
             For example, if the local path is a directory, the remote path's
             base filename will be added onto it (so ``get('foo/bar/file.txt',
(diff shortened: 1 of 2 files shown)


karlcow:
@anarcat do you want me to do one commit?

anarcat:
> @anarcat do you want me to do one commit?

that would be preferable yes

karlcow:
ok done.

davidjmemmett:
Hi @karlcow,

Whilst the comment at the top of [fabric/transfer.py](https://github.com/fabric/fabric/blob/35d7662ee020e8de236577a17571f1428c102479/fabric/transfer.py#L2) is kinda wrong (SCP isn't yet implemented), the other docstrings are still accurately describing the respective arguments as being similar in behaviour to `scp`.

Cheers,
David

davidjmemmett:
In addition, this PR is against `2.0`, which is an older branch of fabric. Closing for now - please feel free to re-open a PR against the `master` branch if you feel strongly about this.

karlcow:
The PR was valid at the time, but is not anymore, because the docstrings have changed. :)

PR #2038, opened 2019-11-26
Implement file tranfers on group class

### LINKED ISSUE #2037: Feature request group transfers
I'd like to be able to use file transfers in `Groups` currently this is not possible because the `Transfer` class requires a `Connection`.

I'll write up a PR.

obeleh:
fixes #2037 

Why?
I'd like to do tranfers to multiple hosts at the same time

What does this PR Do?

It moves the Parallelization for the group classes to a function:
```python
def _operation(self, operation *args, **kwargs):
    pass
```

This operation function is then in charge of executing the operation on all connections. All the methods can now be added to the `Group` class since all those implementations do essentially the same thing: 

Example from the `thread_worker` function:
```python
def thread_worker(cxn, operation, queue, args, kwargs):
    func = getattr(cxn, operation)
    result = func(*args, **kwargs)
    # TODO: namedtuple or attrs object?
    queue.put((cxn, result))
``` 

Example from the `SerialGroup`:
```
func = getattr(cxn, operation)
results[cxn] = func(*args, **kwargs)
```

I'll be making a separate PR for a `GroupTransfer` class so that you can choose whether or not to use that.

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -880,3 +880,6 @@ class Connection(Context):
             self.transport.cancel_port_forward(
                 address=remote_host, port=remote_port
             )
+
+    def is_remote_dir(self, path):
+        return Transfer(self).is_remote_dir(path)
--- a/fabric/group.py
+++ b/fabric/group.py
@@ -105,7 +105,7 @@ class Group(list):
         # TODO: or keep going w/ a "return or except", but the object is
         # largely similar (if not identical) in both situations, with the
         # exception just being the signal that Shit Broke?
-        raise NotImplementedError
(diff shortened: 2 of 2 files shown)


wtayyeb:
take a look at https://github.com/fabric/fabric/pull/2042, I have add group.call(function, ...) that let generally do anything over groups
for your specific problem: `group.call(lambda c: c.put(...))` 
or if you comfort with defining a dedicated def for it will be

```python
def file_transfer_function(c):
   ...

group.call(file_transfer_function)
```
also you could pass `*args, **kwargs` as your function signature.

my PR work on `SerialGroup` and `ThreadingGroup`.

PR #2039, opened 2019-11-26
Add Group transfer class

obeleh:
This builds upon https://github.com/fabric/fabric/pull/2038

But since I wasn't sure if you'd want to have this class as wel, I've moved it to a separate pr


obeleh:
Fixes #2037

PR #2044, opened 2019-12-18
Added support for Paramiko SFTP callbacks

byt3rr:
Just added a callback argument, which is passed on to the built in callback argument in Paramiko's SFTP class. I was working on a script that transfers some pretty heavy files, and so I wanted to show the transfer progress, which this callback is extremely useful for.


hadim:
Any chance to add a similar arg to `put()`? Can you also provide an example of how you show the progress of the transfer?

PR #2072, opened 2020-03-18
customizable host key policy

### LINKED ISSUE #2071: default paramiko host key policy is insecure
I was quite surprised to find that code in Fabric's `Connection` class:

```
        #: The `paramiko.client.SSHClient` instance this connection wraps.
        client = SSHClient()
        client.set_missing_host_key_policy(AutoAddPolicy())
        self.client = client
```

This changes the default `connect` policy from `RejectPolicy()` to `AutoAddPolicy` which means that a ssh connexion will *not* warn when an SSH key changes or on new keys. I find this behavior questionable, especially since it's actually very hard to change, because it's embedded deep inside Fabric (inside the `Connection` constructor).

I tried to monkeypatch this away in a a `prototype.py` but could not: make this work:

```
#!/usr/bin/python3

import fabric
import fabric.connection

class SaferConnection(Connection):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.client.set_missing_host_key_policy(RejectPolicy())

fabric.Connection = SaferConnection
fabric.connection.Connection = SaferConnection

c = fabric.Connection('root@88.99.194.57')

c.run('uptime')
```

for some weird reason, this crashes with:

```
anarcat@curie:tsa-misc(master)$ ./spike.py 
  File "./spike.py", line 10, in <module>
    c = fabric.Connection('root@88.99.194.57')
  File "/home/anarcat/src/tor/tsa-misc/fabric_tpa/__init__.py", line 56, in __init__
    super().__init__(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/fabric/connection.py", line 371, in __init__
    super(Connection, self).__init__(config=config)
TypeError: __init__() missing 1 required positional argument: 'host'
```

It looks like the pre-py3 `super()` call is causing problems here. I tried to fix that in cd004d40 but it seems it is a recurring problem in the code and, besides, it's hardly an intuitive way to make a change.

Shouldn't the default here be safer? Maybe have a prompt like SSH does instead of just gobbling up any SSH key change? I understand if we would want to do TOFU and just trust unknown fingerprints, but from what I can tell, Fabric, by default, will trust *any* SSH server fingerprint whatsoever.

I find this terrifying.

anarcat:
The current Fabric code does not check or save SSH host keys at all. This PR fixes the problem by ensuring we properly call the Paramiko SSH keys loader when we create our connection object. 

We also load the *right* UserKnownHosts file based on the ssh_config, which paramiko does not do out of the box.

We might want to push the latter into paramiko, but first it seems to me the default Fabric policy of "everything goes" should b fixed.

Note that this also allows callers to override the `AutoAddPolicy` that is currently in use.

This should fix #2071

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ script:
   # NOTE: this also runs as the sudo-capable user, even if it's not necessarily
   # doing any sudo'ing itself - the sudo-capable user is also the ssh-able
   # user...
-  - inv travis.sudo-run "inv integration --capture=no"
+  - inv travis.sudo-run "inv integration"
   # Websites build OK? (Not on PyPy3, Sphinx is all "who the hell are you?" =/
   - "if [[ $TRAVIS_PYTHON_VERSION != 'pypy3' ]]; then inv sites www.doctest docs.doctest; fi"
   # Did we break setup.py?
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,10 @@ matrix:
   # pypy3 (as of 2.4.0) has a wacky arity issue in its source loader. Allow it
   # to fail until we can test on, and require, PyPy3.3+. See
   # pyinvoke/invoke#358.
+  # NOTE: both pypy flavors are weirdly unstable on Travis nowadays, even
+  # pre-test-run.
   allow_failures:
+    - python: pypy
     - python: pypy3
   # Disabled per https://github.com/travis-ci/travis-ci/issues/1696
   # fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,7 @@ script:
   - rm -rf tmp
   - pip uninstall -y fabric
   - "PACKAGE_AS_FABRIC2=yes inv travis.test-packaging --package=fabric2 --sanity=\"fab2 --version\""
+  - inv sanity-test-from-v1
 after_success:
   # Upload coverage data to codecov
   - codecov
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2019 Jeff Forcier.
+Copyright (c) 2020 Jeff Forcier.
 All rights reserved.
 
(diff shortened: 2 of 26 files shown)


anarcat:
at this point, I'm stuck on unit tests. i have absolutely no idea how to mock an SSH connexion the way that's required here.. it looks like everything in the current `tests/connection.py` test suite fakes SSH sessions at a much higher level than what we would need here.

so i'm a bit at a loss on where to do the complete patch here, help would be very much welcome on that!

anarcat:
i got as far as this for the unit tests, which is "not very":

```diff
modified   tests/connection.py
@@ -12,6 +12,7 @@
 from mock import patch, Mock, call, ANY
 from paramiko.client import SSHClient, AutoAddPolicy
 from paramiko import SSHConfig
+from paramiko.ssh_exception import BadHostKeyException
 import pytest  # for mark
 from pytest import skip, param
 from pytest_relaxed import raises
@@ -1154,3 +1155,10 @@ def tunnel_errors_bubble_up(self):
 
         def listener_errors_bubble_up(self):
             skip()
+
+    class known_hosts:
+        @raises(BadHostKeyException)
+        def changed_host_key_fails(self):
+            c = Connection('localhost')
+            c.open()
+            
```

anarcat:
the test suite fails, but that's because sphinx cannot reach pyinvoke.org, i doubt it's related to this patch:

```
loading intersphinx inventory from http://pyinvoke.org/objects.inv...
[...]
sphinx.errors.SphinxWarning: failed to reach any of the inventories with the following issues:
```

https://travis-ci.org/github/fabric/fabric/jobs/664111778

... not sure what's up with that, but it's true i'm having trouble reaching pyinvoke.org myself

anarcat:
#2073 should fix that particular build issue.

anarcat:
rebased on top of #2073 because that actually fixes the test suite

anarcat:
this needs to be republished against the master branch, sorry for the noise. :(

PR #2073, opened 2020-03-18
fix intersphinx URLs

anarcat:
It looks like http://pyinvoke.org/ does not respond. Assume the
correct URL is www.pyinvoke.org instead, which does.

And while we're here, switch all of those to HTTPS because we're in
the future now and HTTPS just works.


anarcat:
closing in favor of #1960 because it's more complete.

Issue #1929, opened 2019-01-15
Change default branch to `master`


rectalogic:
It appears fabrics default branch is `2.0` which is far behind master. So any fresh clone is pretty old. Is there a reason not to have `master` be the default branch?

```sh-session
$ git clone https://github.com/fabric/fabric.git
Cloning into 'fabric'...
remote: Enumerating objects: 24900, done.
remote: Total 24900 (delta 0), reused 0 (delta 0), pack-reused 24900
Receiving objects: 100% (24900/24900), 18.40 MiB | 10.16 MiB/s, done.
Resolving deltas: 100% (16709/16709), done.
$ cd fabric/
$ git branch
* 2.0
```

anarcat:
i was also bitten by this when building PRs, which i mistakenly rolled against 2.0. it was fairly easy to catch up (only one commit had a conflict) but it would seem wise to follow "master" here instead of a release branch.

PR #2076, opened 2020-03-19
customizable host key policy

### LINKED ISSUE #2071: default paramiko host key policy is insecure

anarcat:
The current Fabric code does not check or save SSH host keys at all. This PR fixes the problem by ensuring we properly call the Paramiko SSH keys loader when we create our connection object.

We also load the right UserKnownHosts file based on the ssh_config, which paramiko does not do out of the box.

We might want to push the latter into paramiko, but first it seems to me the default Fabric policy of "everything goes" should b fixed.

Note that this also allows callers to override the AutoAddPolicy that is currently in use.

This should fix #2071

This is a rebase of #2072 on top of master (instead of 2.0, see #1929 for that) and with #1960 merged (instead of #2073) to fix the docs.

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -9,7 +9,7 @@ from decorator import decorator
 from invoke import Context
 from invoke.exceptions import ThreadException
 from paramiko.agent import AgentRequestHandler
-from paramiko.client import SSHClient, AutoAddPolicy
+from paramiko.client import SSHClient, RejectPolicy
 from paramiko.config import SSHConfig
 from paramiko.proxy import ProxyCommand
 
@@ -142,7 +142,7 @@ class Connection(Context):
     transport = None
     _sftp = None
     _agent_handler = None
-    default_host_key_policy = AutoAddPolicy
+    default_host_key_policy = RejectPolicy
 
     @classmethod
     def from_v1(cls, env, **kwargs):
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -457,18 +457,8 @@ class Connection(Context):
         self.connect_kwargs = self.resolve_connect_kwargs(connect_kwargs)
 
         #: The `paramiko.client.SSHClient` instance this connection wraps.
-        client = SSHClient()
-        if self.default_host_key_policy is not None:
-            logging.debug('host key policy: %s', self.default_host_key_policy)
-            client.set_missing_host_key_policy(self.default_host_key_policy())
-        known_hosts = self.ssh_config.get('UserKnownHostsFile'.lower(),
-                                          '~/.ssh/known_hosts')
-        logging.debug('loading host keys from %s', known_hosts)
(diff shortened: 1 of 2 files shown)


d1b:
cc @bitprophet . 

As an aside I think it is well overdue for fabric to offer a sensible & secure default for users and not blindly auto-add ssh keys. Of course changing the default behaviour would require a new major (if following semver) version bump.

tacerus:
Thank you for the patch, upon trying to apply this on top of our distribution package, I noticed the test suite to fail due to still missing a minor adjustment which I added here: https://github.com/tacerus/fabric/commit/9c915064babc30daed6ea22431ab02d3626d5d3c (I suppose it could alternatively be read from `default_host_key_policy` now).

tacerus:
It seems some Python 3.6 tests fail here in the upstream pipeline now as a result, sorry about that, I did not test with 3.6. :-(

Issue #2085, opened 2020-05-04
Documentation - key_filename example is incorrect in connection.py


riazjahangir:
Documentation error:

In the `Connection` API docstring (https://docs.fabfile.org/en/latest/api/connection.html#fabric.connection.Connection), the example code is:

```python
c = Connection(
    host="hostname",
    user="admin",
    connect_kwargs={
        "key_filename": "/home/myuser/.ssh/private.key",
    },
)
```

However running this throws an error:
```python
AttributeError: 'str' object has no attribute 'extend'
```

The `key_filename` expects a list, so the example should read:

```python
c = Connection(
    host="hostname",
    user="admin",
    connect_kwargs={
        "key_filename": ["/home/myuser/.ssh/private.key"],
    },
)
```

DhawalKhapre:
Can I take up this issue. I'm a beginner.
I just have to change the example on the webpage right?

davidjmemmett:
Hi @riazjahangir,

The example given is correct - `key_filename` can either be a string or a list - please [see the appropriate documentation for the connect_kwargs](http://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient). If this is still an issue, please feel to answer the questions below to give us a better idea of what is happening.


**What to put in your bug report**

Make sure your report gets the attention it deserves: bug reports with missing information may be ignored or punted back to you, delaying a fix. The below constitutes a bare minimum; more info is almost always better:

- What version of the Python interpreter are you using? Are you using an alternative interpreter such as PyPy?
- What operating system are you using both client & server-side?
- Are you using OpenSSH server or something else?
- Which version or versions of the software are you using?
- How can the developers recreate the bug on their end? If possible, include a copy of your code, the command you used to invoke it, and the full output of your run (if applicable.)
A common tactic is to pare down your code until a simple (but still bug-causing) “base case” remains. Not only can this help you identify problems which aren’t real bugs, but it means the developer can get to fixing the bug faster.

Cheers,
David

davidjmemmett:
Hah, please ignore my previous comment - I should have tested the issue before speaking!

Example fabfile:
```python
from fabric import Connection

c = Connection(
    host="localhost",
    user="blah",
    connect_kwargs={
        "key_filename": "/home/blah/.ssh/private.key",
    },
)
```

Traceback:
```
$ fab
Traceback (most recent call last):
  File "/home/david/Dropbox/Projects/fabtest/env/bin/fab", line 11, in <module>
[...]

PR #2115, opened 2020-10-02
Solving issue #2085

DhawalKhapre:
I have made the required corrections for issue #2085


davidjmemmett:
As per the comment I just left on #2085, I don't _believe_ this is an active issue, it's possibly a version mismatch somewhere. Closing as this MR is against an old branch.

Issue #2142, opened 2021-02-19
WinError87 when executing local on Connection


KantiCodes:
Hi all, I`m new to Invoke and even newer to Fabric!

I have some functionality in Invoke and now I`m trying to make a similar demo using **fabric**

Following code results in WinError87
![image](https://user-images.githubusercontent.com/24612001/108508418-14019c00-72bc-11eb-8dc5-055ab9f94fd3.png)

![image](https://user-images.githubusercontent.com/24612001/108508478-2bd92000-72bc-11eb-8abf-192437f95790.png)
[WinError 87] Parameter is incorrect

I know it's probably caused by me misunderstanding something - but I`d really appreciate a hint :)

Setup:

Python 3.8.5
Fabric 2.6.0

davidjmemmett:
Hi KantiCodes,

Just to confirm, you have an SSH server running on localhost on port 8050, and the username you're wanting to connect to is 45502?

Which SSH server are you using? If you open an SSH connection using the standard openssh client (`ssh 45502@localhost -p8050`) and run `cd ..`, does this work as expected?

Cheers,
David

KantiCodes:
Hi @davidjmemmett 

Thanks so much for taking time to look at my problem :)

No, I do not have an SSH running - my understanding was that until I only run c.local() the connection is not established and it's just as if I was using invoke.

Sorry about that.

To add a little background to what I want to do.

I created invoke module and wrapped it into my own pip installing library: http://docs.pyinvoke.org/en/stable/concepts/library.html#reusing-invoke-s-cli-module-as-a-distinct-binary

Here is a simplified example:

tasker/main.py
![image](https://user-images.githubusercontent.com/24612001/108511244-f0d8eb80-72bf-11eb-9615-17616d5a194d.png)

tasker/tasks.py
![image](https://user-images.githubusercontent.com/24612001/108511278-fc2c1700-72bf-11eb-99ea-42ec78696e45.png)

I can install and access my wrapped "module"

![image](https://user-images.githubusercontent.com/24612001/108511451-385f7780-72c0-11eb-89b9-1369ac86c827.png)

Now my goal:
- For some of the tasks I need to communicate with web server, therefore I would like to use Fabric and keep my current(Invoke's) functionality and add new tasks that.

As far as I understand @bitprophet designed fabric to be **Invoke +** therefore I would like to keep my Invoke solution and add being able to connect over SSH to web servers.

To wrap up:
[...]

Issue #1752, opened 2018-05-14
Separation of local vs remote lowercase-C contexts (for cd, prefix, run/local, etc)


bitprophet:
(Cannot find an old ticket for this :( thought there was one...)

Fabric 1 had `with cd:` and `with lcd:` and similar dual APIs for users that needed to track context-y state (cwd, prefixes list etc) differently for the remote vs local ends.

Invoke only has the one local context and doesn't care.

Fabric 2 inherits from Invoke but failed to grow distinct remote-vs-local APIs before 2.0 (besides the much earlier implementation of "move Invoke's `run` to Fabric 2's `local`", which presents its own set of problems around "what kind of Context am I dealing with right now and what methods does it have?").

This needs solving and I assume in the obvious fashion, i.e. growing new API members (named explicitly either after local or remote) and the appropriate state storage (on Connection or in config).

This will sadly be a backwards compatibility break (insofar as `with c.cd()` currently changes both ends and will revert to changing only one) but the software's only been out a week and I think we can argue that this was a bug. Otherwise it'll be the shortest lived major version number ever 😐 

This is strongly related to #1686 which is for the run-vs-local config divergence.

rpkilby:
Hi @bitprophet - were you thinking of #1637?

bitprophet:
#1637 got rebased into #1755, FTR. And I still have to doublecheck whether that approach works for me or not!

revolter:
So there is currently no workaround for `lcd`?

bitprophet:
#1858 brings up one possible angle on the run-vs-local side of things, namely adding a `local` to `invoke.Context` so that the two APIs resemble one another more closely.

I think we still want to reject this approach, as it sullies Invoke and doesn't entirely solve the issue (we're still gonna have problems with things like `cd`, or any new API members either side of the equation adds later).

I also do still think (at this point in time, months later) that #1755's approach may be the only sensible way forward, tho I have to reread it and give a harder think on any other obvious APIs that come to mind. But having the two sides of the conversation as discrete API objects seems hard to beat for explicitness. It's just gonna be hard to wrangle re: backwards compatibility (& if put out in a backwards compat fashion it'll introduce extra confusion; however given there's already a lot of confusion around this......)

rpkilby:
[...]

PR #2144, opened 2021-02-23
Drop replace_env from global-scope config

davidjmemmett:
There may be other implications that I'm unaware of, though from my perspective the tests pass and the issue appears to be resolved.

I've removed `run.replace_env` from Fabric's config override, and replaced it with a kwarg override on `fabric.runners.Remote.run`. In some version of Python since 3.6.6, `_winapi.CreateProcess` starts throwing the error `OSError: [WinError 87] The parameter is incorrect` due to missing environment variables. Ensuring that Fabric doesn't interfere with Invoke's local calls (by removing said globally-scoped configuration variables) seems like the right way to achieve this.

Issue: fabric/fabric#2142
See also: 4358f344e5d6b59dc99d954ad8287b4021d0ed81

### DIFF
--- a/fabric/config.py
+++ b/fabric/config.py
@@ -308,7 +308,6 @@ class Config(InvokeConfig):
             "inline_ssh_env": False,
             "load_ssh_configs": True,
             "port": 22,
-            "run": {"replace_env": True},
             "runners": {"remote": Remote},
             "ssh_config_path": None,
             "tasks": {"collection_name": "fabfile"},
--- a/fabric/runners.py
+++ b/fabric/runners.py
@@ -58,7 +58,7 @@ class Remote(Runner):
         self.channel.exec_command(command)
     
     def run(self, command, **kwargs):
-        kwargs['replace_env'] = True
+        kwargs.setdefault('replace_env', True)
         return Runner.run(self, command, **kwargs)
     
     def read_proc_stdout(self, num_bytes):
--- a/fabric/runners.py
+++ b/fabric/runners.py
@@ -56,7 +56,11 @@ class Remote(Runner):
             else:
                 self.channel.update_environment(env)
         self.channel.exec_command(command)
-
+    
+    def run(self, command, **kwargs):
+        kwargs['replace_env'] = True
+        return Runner.run(self, command, **kwargs)
+    
     def read_proc_stdout(self, num_bytes):
         return self.channel.recv(num_bytes)
 
--- a/sites/docs/concepts/configuration.rst
+++ b/sites/docs/concepts/configuration.rst
@@ -46,9 +46,9 @@ Default configuration values
 Overrides of Invoke-level defaults
 ----------------------------------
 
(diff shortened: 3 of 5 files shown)


davidjmemmett:
Travis CI builds appear to be broken due to an old version of pip.

davidjmemmett:
Accidentally deleted the wrong branch in my repo - re-opened.

bitprophet:
Ah, this arguably falls under #1752 insofar as I think the 100% "correct" fix here is for Fabric to have a `local.*` settings tree for its "local run" (or, depending, a `remote.*` one...hmm).

I'm wary of having the perfect delay the good, and I don't see any obvious reasons why using this as a temporary workaround would be actively harmful, so I'm tempted to merge. I think we need a few more changes tho:

- the kwarg override should be "gentle" and use `kwargs.setdefault` in case a user is explicitly giving it for some reason
- there ought to be another test proving that this kwarg override is happening (currently the test change only proves that the _config_ override is _not_ happening) and also proving the setdefault is working (caller can override)
- I am relatively sure we have a note about the `run.replace_env` override in the docs, which needs tweaking (probably to say that the change is now happening at the keyword argument level)

LMK if you'd rather I handle some or all of that but for now I figure this is good practice for Learning How Jeff Thinks 🙃

davidjmemmett:
I _think_ the documentation part is still fine - updating to say "defaults to False" if that helps?

davidjmemmett:
I think I'll leave the changelog part for you for now - not just because I haven't used reStructuredText before, but I don't know what version will be next. I can make an ?

davidjmemmett:
Hi @bitprophet,

Regarding your feedback (also apologies for the lazy commit messages - they do describe my intent at least):

- [x] the kwarg override should be "gentle" and use kwargs.setdefault in case a user is explicitly giving it for some reason
- [x] there ought to be another test proving that this kwarg override is happening (currently the test change only proves that the config override is not happening) and also proving the setdefault is working (caller can override)
- [ ] I am relatively sure we have a note about the run.replace_env override in the docs, which needs tweaking (probably to say that the change is now happening at the keyword argument level)
- [ ] Changelog entry! (I seriously need to add a check for this to my CI, maybe I'll find time during the Circle work that is ongoing...)

I'm not 100% sure what needs doing regarding the last two items.

Cheers,
David

bitprophet:
Pushing some commits and a merge for this - check the commits to see how I handled the docs and changelog.

The changelog specifically is actually designed to avoid needing to know what release is next - it's basically a stream/log of changes which my overengineered Sphinx changelog plugin (Releases) parses out into release sections at render time. So any given change we make just wants to add one new bullet list item at the top of the list, no matter what else is going on or what branch it's in.

bitprophet:
Had to do a bit of rebasing for things like black formatting so this won't show merge cleanly - closing.

bitprophet:
Ended up with one relevant new commit - 35d7662ee020e8de236577a17571f1428c102479

PR #2145, opened 2021-02-23
Master ignore eclipse gubbings

davidjmemmett:
Just a nicety for me - if you wouldn't mind :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ htmlcov
 env/
 .project
 .pydevproject
+.settings/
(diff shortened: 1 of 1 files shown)


bitprophet:
Gah I'm always so damn conflicted on these things. The degraded case is that the .gitignore is literally a hundred lines long with every single editor and tool's detritus listed; the last few years I've been more likely to remove lines here than add them.

Is there a reason this doesn't work better in your `~/.config/git/ignore`? If it truly makes your life easier I'll do it but a lot of the time I assume this means folks aren't aware they can do it on a per user basis.

davidjmemmett:
I’ve been using git for many years and I didn’t know that file exists! Yup, I’m going to do that.

PR #2123, opened 2020-10-28
clarify documentation for Connection.put

dsch:
Fix the wrong documentation for Connection.put method.

PR #2154, opened 2021-03-20
clarify documentation for Connection.put

davidjmemmett:
New pull request against `master` branch.
Original pull request from @dsch: https://github.com/fabric/fabric/pull/2123

> Fix the wrong documentation for Connection.put method.


bitprophet:
Actually in this case, doc fixes tend to want to be merged to old branches and merged-up; this way the docs for the old versions get the fix. I'll cherry pick and merge-up (it's not something github can do automatically).

PR #2155, opened 2021-03-25
Local include std streams in error

timsnyder:
Whenever I hit an error during `rsync_project()` or `local()`, I'm always annoyed that the stdout and stderr of the command isn't included in the output from fabric.  It forces me to copy/paste the failing command out of the log and rerun it to see what went wrong.  I finally looked at the Fabric 1.14.0 code to see why this is happening. [rsync_project() just ends up using `local()`,](https://github.com/fabric/fabric/blob/1.14.0/fabric/contrib/project.py#L161)  so I'll continue to describe the issue for local() and stop whining about `rsync_project()` but the same behavior is observed there.

When `local(capture=True)` hits a returncode that isn't ok, [it calls error(), passing in the captured stdout and stderr](https://github.com/fabric/fabric/blob/1.14.0/fabric/operations.py#L1239).  However, `error()` [only adds the stdout and stderr to the error message if fabric is configured to `not fabric.state.output.std{out|err}`](https://github.com/fabric/fabric/blob/1.14.0/fabric/utils.py#L354-L357).  That makes no sense when only looking at `local(capture=True)` but when you look at how `run()` calls `error()` it does because `run()` will have already printed the stdout and stderr of the command, obeying `fabric.state.output.std{out|err}`.  Given that `local(capture=True)` will not have printed the stdout and stderr of the command, the call to `error()` from `local(capture=True)` should contextually invert the sense of `fabric.state.output.std{out|err}` so that the error message will include stdout and stderr that is captured and passed to `error()`.

### DIFF
--- a/fabric/operations.py
+++ b/fabric/operations.py
@@ -1236,7 +1236,12 @@ def local(command, capture=False, shell=None):
     if p.returncode not in env.ok_ret_codes:
         out.failed = True
         msg = "local() encountered an error (return code %s) while executing '%s'" % (p.returncode, command)
-        error(message=msg, stdout=out, stderr=err)
+        # error()'s logic for appending the stdout and stderr to message assumes the caller has already printed
+        # them if they are to-be shown.  While this assumption is correct for run(), it is not for local(capture=True).
+        # So, if capture=True, we temporarily hide stdout and stderr so that error() will think they should
+        # be included in the error message.
+        with settings(hide('stdout', 'stderr') if capture else None):
+            error(message=msg, stdout=out, stderr=err)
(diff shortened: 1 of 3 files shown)


davidjmemmett:
Hi @timsnyder,

Thank you for the pull request, but I'm afraid that fabric v1 is no longer actively receiving new features/bugfixes; new releases will only be made for urgent security vulnerabilities. We actively encourage all users to [upgrade to fabric v2](http://www.fabfile.org/upgrading.html).

Cheers,
David

timsnyder:
Thanks for the reply to the PR.  I completely understand your response. :)

Issue #2026, opened 2019-10-18
gateway connection not respect timeout


dhruvinsh:
Hi,
Below code snippet shows the servers I have. host1 act as a jump server to connect host2. 
``` python
host1 = '10.254.90.25'
host2 = 'blade'
with Conection(host=host1, connect_timeout=10) as c:
    with Connection(host=host2, gateway=c, connect_timeout=10) as b:
        b.run('uname -r')
```
But lets say host2 is not available, then this code get stuck for at lease 3 minutes.
inspecting source code I found below in open_gateway method,
```python
        return self.gateway.transport.open_channel(
            kind="direct-tcpip",
            dest_addr=(self.host, int(self.port)),
            # NOTE: src_addr needs to be 'empty but not None' values to
            # correctly encode into a network message. Theoretically Paramiko
            # could auto-interpret None sometime & save us the trouble.
            src_addr=("", 0),
        )
```

@bitprophet  why here `open_channel` is missing timeout arg? by default it will take 3600 seconds if its None. is there known reason for this behavior?

IMO, it should be like below
```python
        return self.gateway.transport.open_channel(
            kind="direct-tcpip",
            dest_addr=(self.host, int(self.port)),
            # NOTE: src_addr needs to be 'empty but not None' values to
            # correctly encode into a network message. Theoretically Paramiko
            # could auto-interpret None sometime & save us the trouble.
            src_addr=("", 0),
            timeout=self.connect_timeout
        )
```

dhruvinsh:
Pull request submitted, hope that gets accepted. :)

PR #2158, opened 2021-03-30
Have open_gateway honor connect_timeout for gateway channel creation

jmeichle:
Without the ability to specify this timeout, behavior can be quite frustrating when your gateway host is reachable but the target destination is not reachable. An example use case spinning up a new EC2 instance and waiting for SSH to stabilize.

I saw the comment about how to best handle timeouts and figured that an explicit kwarg to Connection for the gateway timeout made the most sense, and that it is only used on connection objects that are used for gateways.

If there are any other places to update / tests to run, let me know.

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -665,8 +665,6 @@ class Connection(Context):
         # needs...
         # TODO: and the inverse? allow users to supply their own socket/like
         # object they got via $WHEREEVER?
-        # TODO: how best to expose timeout param? reuse general connection
-        # timeout from config?
         return self.gateway.transport.open_channel(
             kind="direct-tcpip",
             dest_addr=(self.host, int(self.port)),
@@ -674,6 +672,7 @@ class Connection(Context):
             # correctly encode into a network message. Theoretically Paramiko
             # could auto-interpret None sometime & save us the trouble.
             src_addr=("", 0),
+            timeout=self.connect_timeout,
         )
 
     def close(self):
--- a/tests/connection.py
+++ b/tests/connection.py
@@ -756,6 +756,36 @@ class Connection_:
             kwargs = open_channel.call_args[1]
             assert kwargs["kind"] == "direct-tcpip"
             assert kwargs["dest_addr"], "host" == 22
+            assert kwargs["timeout"] is None
+            # Expect result of that channel open as sock arg to connect()
+            sock_arg = mock_main.connect.call_args[1]["sock"]
+            assert sock_arg is open_channel.return_value
+
+        # NOTE: does more involved stuff so can't use "client" fixture
+        @patch("fabric.connection.SSHClient")
+        def uses_timeout_on_the_gateway_io(self, Client):
+            "uses Connection gateway as 'sock' arg to SSHClient.connect"
+            # Setup
+            mock_gw = Mock()
+            mock_main = Mock()
+            Client.side_effect = [mock_gw, mock_main]
+            # connect_timeout=7 here to be explicit and to
+            # distinguish from the timeout for the main connection
+            gw = Connection("otherhost", connect_timeout=7)
+            gw.open = Mock(wraps=gw.open)
+            main = Connection("host", gateway=gw, connect_timeout=12)
+            main.open()
+            # Expect gateway is also open()'d
+            gw.open.assert_called_once_with()
+
+            # Expect direct-tcpip channel open on 1st client
+            open_channel = mock_gw.get_transport.return_value.open_channel
+            kwargs = open_channel.call_args[1]
+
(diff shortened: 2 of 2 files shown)


bitprophet:
Thanks! Very glad to see docstring and tests included 👍🏻 

I still wonder, re: the TODO comment, if using the outer/destination Connection object's `connect_timeout` would make more sense? Thinking about it from first principles right now:

- It feels intuitive - the user is already encouraged to think about the outer Connection as the "main" Connection, with the gateway being an "implementation detail" of sorts.
- It means that connecting to the outer Connection target uses `connect_timeout` regardless of whether a gateway is in play (eg say your network topology changes or is in flux).
- Conversely, if we use a new `gateway_timeout` as implemented here, the outer Connection's `connect_timeout` is actually _ignored_, which could be surprising
  - _and_ it's only ignored when the gateway's being used, so if again you end up with a temporary network change, suddenly a different setting is being used.
- It's a bit less code and fewer options/kwargs, which is always ideal _if_ we're not painting ourselves into a corner (which I don't think we would be here?)
- On top of that, if we imagine multi-hop setups, juggling two kwargs instead of one starts to get even harder to reason about.

To make this really clear, what I'm proposing is eg:
``` python
gateway = Connection('bastion', connect_timeout=5)
real_target = Connection('slowboat', connect_timeout=30, gateway=gateway)
```
which would have the following properties:
- Can twiddle the timeout of either network link independently (and this scales up to N hops as well as anything does)
- The connection from `localhost` to `gateway` has a 5s timeout (fast host)
- The connection from `gateway` to `slowboat` has a 30s timeout (per your described scenario)
- If we nixed the `gateway=gateway` kwarg from `real_target` (eg say you had some awful problem with the bastion host and temporarily lifted firewall restrictions), no other change is necessary, you still have a 30s connect timeout directly talking to `slowboat`.

Apologies for the verbosity! Not trying to slam your initial feeling for the shape of the feature, I might've done the same initially. Open to counterarguments 😁

jmeichle:
@bitprophet :wave:

Your proposed changes sound fine to me. I typically prefer to be explicit with separate kwargs for different functionality, but can totally understand the benefits of having `connect_timeout` both mean "timeout for the connection to the jump host" and "timeout of gateway channel opening over the existing connection" from an end user perspective even if they are technically separate operations.

The background for this PR is I am using these fabric connection objects for an infrastructure lifecycle management. Connections to the proxy/jump gateway host are quite reliable as that infrastructure is long lived/monitored/etc, but connections via the gateway channel might not be stable (think: polling loop until SSH stabalizes for a host that is launching and might not be online yet). Without this PR fabric relied on a default timeout value (I think from paramiko) which is *very* long (or no timeout), and having the ability to define a really low gateway timeout seemed useful even if the main connection to the proxy host should have a long timeout. 

I'll push an update later today, and otherwise will require some advise on how to handle the travis failures as they seem unrelated to this change set (dependency issues, also travis is setup to use ubuntu trusty which is EOL). If desired, I can attempt to update travis.yml to use a non-EOL ubuntu.

bitprophet:
@jmeichle great! And don't worry about Travis for now, they're going down the tubes and I am midway through porting things to CircleCI. For now, as long as the tests pass for both of us locally that'll do for now. (My workstation tests Python 2+3 at least.)

dhruvinsh:
I did not notice this pull request, and submitted new one just today. https://github.com/fabric/fabric/pull/2172

I reported the issue here: https://github.com/fabric/fabric/issues/2026 with solution to use existing connect_timeout parameter. Which mitigate the use of additional argument.
```
proxy = Connection("gateway-server", connect_timeout=10)
server = Connection("main-server", gateway=proxy, connect_timeout=1)
```

Benefit of this approach would be single argument handles have same and expected behaviour.

jmeichle:
@bitprophet I think i have this updated correctly, sorry for the delays. I might also have it backwards. Would appreciate eyes on the test since its buried pretty deep and i want to make sure im not doing something silly (as it was a copy-paste of another low level test).

davidjmemmett:
I think you might have wanted to rebase rather than merge from `master`, as it shows commits from the main trunk in your merge request.

davidjmemmett:
I'll prod @bitprophet anyway

jmeichle:
Sorry, I'll squash this branch

jmeichle:
@davidjmemmett squashed to one commit after merging master (I have a weird local git workflow where I dislike rebase and always merge+squash)

dhruvinsh:
@bitprophet  sorry to tag you, but if its okay, can we please merge this?

Philimanjar0:
Can this merge and make it into a release? This change would be very useful to me.

PR #2168, opened 2021-06-03
Use standard library unittest.mock on Python 3.6 and later

musicinmybrain:
Drop the dependency on PyPI `mock` beginning with Python 3.6, as it is entirely replaced by `unittest.mock` in the standard library.

While `unittest.mock` appeared in Python 3.3, we explicitly check that the Python version is at least 3.6 so that the `assert_called()` and `assert_called_once()` methods are available on the `Mock` class.


musicinmybrain:
There are CI failures for reasons that appear unrelated to this PR.

PR #2172, opened 2021-06-24
proxy gateway supports timeout #2026

dhruvinsh:
If the targeted host is not available while using gateway, connection would stall for 2 min. Connection object with timeout parameter and valid gateway object, It does not respect the timeout. This commit try to solve the underlying issue.


bitprophet:
Let's consider this covered by #2158 as long as the proposed changes in that discussion land. Thx!

PR #2174, opened 2021-07-27
refactor: implement absolute imports

kuwv:
This PR is intended to help improve the security of applications using fabric by using explicit paths.

Relative paths can potentially be exploited my malicious libraries. This coupled with the potential use of sudo capabilities can allow for privilege escalation.

This is due to the module cache being world writable within CPython.


kuwv:
I'll close this and put together an example.

PR #2179, opened 2021-10-11
yield the sock in foward_local

graingert:
also open and close the foward_local server socket on the main thread


davidjmemmett:
Hi @graingert - whilst we certainly appreciate contributions for new features, would you also be able to provide some tests to support your changes? I appreciate testing sockets & threads isn't the easiest!

PR #2189, opened 2022-01-12
Allow original_host to be used in ProxyJump option

erikzeneco:
Fix infinite recursion that happens when alias/original_host is used.


davidjmemmett:
Nice find! Is there any chance you could write a test to support this change? If not, please shout up and someone can help.

PR #2190, opened 2022-01-19
Document implications of close() behavior

mrclay:
Documents that close() doesn't block until the SSH connection actually is ended. Useful info if your SSHd cannot accept multiple connections and you need to immediately follow fabric commands with a separate connection (e.g. rsync).


davidjmemmett:
Interesting edge-case - I'll assign for approval.

PR #2205, opened 2022-04-04
Merge `fabric3` into v1 mainline

bitprophet:
Per discussion on the mailing list with @mathiasertl and others. See also my [roadmap](https://bitprophet.org/projects/#roadmap).

This is purely for the benefit of users still soldiering away on Python 2 + Fabric 1, who'd rather upgrade their dev/deploy systems to Python 3 first instead of jumping to the Fabric 2 API right away. Our primary focus remains on Fabric 2 (and its constituent libraries, Invoke and Paramiko).

This is expected to be the last set of commits to `v1` unless severe bugs are discovered.


arnimarj:
I'm one of those still toiling away with a mix of `fabric3` and `fabric2`, and I'm aware of one issue with `fabric3`:

```shell
$ fab
  File "/home/arni/.config/dohop/tools_venv3/bin/fab", line 5, in <module>
    from fabric.main import main
  File "/home/arni/.config/dohop/tools_venv3/lib/python3.10/site-packages/fabric/main.py", line 12, in <module>
    from collections import Mapping
ImportError: cannot import name 'Mapping' from 'collections' (/home/arni/.pyenv/versions/3.10.3/lib/python3.10/collections/__init__.py)
```

Changing the import line to `from collections.abc import Mapping` fixes it

bitprophet:
Nice, thanks @arnimarj ! Guess that's because it was officially removed in 3.10 and a test dependency of ours dies at/after 3.9 so I wasn't able to add it to the test matrix. I'll make that fix now...ok, needs a somewhat newer six, unlikely to be an issue. Pushing that fix now.

arnimarj:
Just tried the same my scripts and indeed it works! Thank you for this effort

un33k:
How to get this change in from pypi?

[Use newer six for collections_abc so Python 3.10 might work](https://github.com/fabric/fabric/pull/2205/commits/1f52202999e5857c5997265a38763ab089594e34)

It seems after 8 months, then change is still missing from Fabric3==1.14.post1

PR #2209, opened 2022-05-17
Use sys.executable when testing with -m fabric

s-t-e-v-e-n-k:
The current test case assumes /usr/bin/python is the current running
Python release, which it may not always be, so make use of
sys.executable to re-exec the interpreter.


MrAnderson-1999:
is there still something missing here?
i will be glad to contribute

Issue #2102, opened 2020-07-27
Cannot port forward


nitrocalcite:
I would like to forward some ports from a remote SSH server to my local computer.  Following this docs, I tried this:
```
from fabric import Connection
import time
c = Connection(remote)
with c.forward_local(8000):
    while True:
        time.sleep(1)
```
As a simple test I used `python -m http.server 8000` on the remote machine.  If I use `ssh user@ip -L8000:localhost:8000` I can connect to the remote Python server no problem.  The above script produces no errors but also doesn't forward any traffic.

I'm running `Fabric 2.5.0` on Windows 10.  I see this in `netstat -an`:

```
  Proto  Local Address      Foreign Address  State
  TCP    127.0.0.1:8000     0.0.0.0:0        LISTENING
```
I thought perhaps I misunderstood the semantics of `forward_local`, but creating the `TunnelManager` manually in the manner that `forward_local` does yields the exact same results.

Am I doing something completely wrong, or is this functionality broken somehow?  Any known workarounds?
Thanks!

nitrocalcite:
The problem seems to be in the `TunnelManager` loop (line 63 of `tunnels.py`)
```
while not self.finished.is_set():
    # Main loop-wait: accept connections on the local listener
    # NOTE: EAGAIN means "you're nonblocking and nobody happened to
    # connect at this point in time"
    try:
        tun_sock, local_addr = sock.accept()
        # Set TCP_NODELAY to match OpenSSH's forwarding socket behavior
        tun_sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
    except socket.error as e:
        if e.errno is errno.EAGAIN:
            # TODO: make configurable
            time.sleep(0.01)
            continue
        raise
    ...
```

The `except` clause is not functioning as intended.  When `sock.accept` fails in nonblocking mode, it raises a `BlockingIOError` that is not caught by the existing `except` clause.  

So, the fix is pretty straightforward: add another `except` clause like so:
```
except BlockingIOError:
     time.sleep(0.01)
     continue
```
before the existing `except` clause.  Startup and shutdown of the thread appear to proceed cleanly and traffic is forwarded properly for me.

I'm not sure exactly when this behavior changed, but this is the behavior I'm seeing on Python 3.7.7 on Windows 10.  It's not immediately obvious that this is happening, as the stack trace for this exception isn't printed until the thread is shut down, at which point it's not clear if the exception was shutdown related.

[...]

PR #2221, opened 2022-08-05
forward_local() does not work on windows.

benlec:
Tunnel.py code should handle BlockingIOError happening on windows when try to call forward_local()

More discussed on the below issue reported. @nitrocalcite found the fix, cudos to him
#2102


micahg:
Hey, this is nice -- it works for me (until the function its called from ends)

PR #2242, opened 2023-01-27
Fix: call parent `stop()` on `Remote` runner

### LINKED ISSUE #2241: Fabric 3.0.0 locks main thread until timeout finish
Fabric 3.0.0 prevents the interpreter to exit if a timeout was specified when running a remote command

For example this small script:
```python
import fabric
import sys

conn = fabric.Connection("localhost", user="ubuntu", connect_kwargs={"key_filename":"key"})
sys.exit(conn.run("whoami", hide=True, warn=True, timeout=30)
```
will generate the following output:
```
Command exited with status 0.
=== stdout ===
ubuntu

(no stderr)
```
but the interpreter will hang until the the timeout.
Hitting `ctrl-C` while hanging generates the following traceback:
```
Exception ignored in: <module 'threading' from '/usr/lib/python3.9/threading.py'>
  File "/usr/lib/python3.9/threading.py", line 1428, in _shutdown
    lock.acquire()
KeyboardInterrupt:
```
This happens no matter the server we connect.

This is a special problem since there are some commands that needs a huge timeout windows, and if the commands executes relatively fast, the program won't finish until the timeout is acomplished.

**Expected behaviour**
Program finish inmediatly

**Environment**
Python 3.9
Ubuntu Linux 5.4.0-126-generic #142-Ubuntu
Fabric 3.0.0
Paramiko 3.0.0
Invoke 2.0.0

puerkito66:
This should resolve #2241

### DIFF
--- a/fabric/runners.py
+++ b/fabric/runners.py
@@ -114,7 +114,7 @@ class Remote(Runner):
         return Result(**kwargs)
 
     def stop(self):
-        super()
+        super().stop()
         if hasattr(self, "channel"):
             self.channel.close()
 


v3ss0n:
looks good to me

DamonHolland (approved):

bitprophet:
Committed my own version of this that had tests and changelog, but credited anyways ;) thx!

PR #2243, opened 2023-02-08
Reject ssh unknown ssh keys by default & warn when an unknown one is seen.

d1b:


d1b:
This is effectively untested.

PR #2245, opened 2023-02-20
pass Connection.connect_timeout to paramiko open_channel

brunson:
Working with 3.0.0 yesterday I wasn't able to set timeout on the initial connection. It seems it isn't being passed into paramiko open_channel and adding this one liner give me the correct behavior. 

If this is not the correct fix, let me know and I'll make it then submit a new PR.


dhruvinsh:
Duplicate of #2158 and #2172

Better implementation is already done on #2158 and just waiting to get merge.

PR #2299, opened 2024-04-03
Python 2.7 support (WiP)

SamuelMarks:
Happy to finish this PR; just want to check that you'd accept it or if you're definitely not interested in supporting older versions


davidjmemmett:
Python 2.7 is EOL and thereby unsupported; efforts are much better placed supporting newer releases of Python

PR #2337, opened 2025-06-02
do not break type decorations in fabric tasks

anarcat:
The invoke.task decorator has type decorations which means that decorated functions can still be typed-checked.

But it seems fabric was not converted to that yet, so you get errors like this with mypy --strict and, indeed, those functions are not checked for typing.

fabric_tpa/postgresql.py:505: error: Untyped decorator makes function "backup" untyped  [misc]

The patch reuses the type signature of invoke.task. It's not actually the exact recommendations mypy provides for that kind of stuff:

https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators

... but given the complexity of the decorator, it's not too bad.

### DIFF
--- a/fabric/tasks.py
+++ b/fabric/tasks.py
@@ -1,3 +1,5 @@
+from typing import Callable, Any
+
 import invoke
 
 from .connection import Connection
@@ -21,7 +23,7 @@ class Task(invoke.Task):
         super().__init__(*args, **kwargs)
 
 
-def task(*args, **kwargs):
+def task(*args: Any, **kwargs: Any) -> Callable:
     """
     Wraps/extends Invoke's `@task <invoke.tasks.task>` with extra kwargs.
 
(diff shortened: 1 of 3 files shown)


anarcat:
i managed to pick up some signature that works better!

```python
from typing import TypeVar, Callable, overload, Any

F = TypeVar("F", bound=Callable[..., Any])

@overload
def task[F: Callable[..., Any]](func: F, /) -> F: ...

# Decorator with arguments
@overload
def task[F: Callable[..., Any]](**kwargs: Any) -> Callable[[F], F]: ...
```

that works! i can now decorate my functions... it's really confusing though because i can't just drop those type hints in my code, as if i make a `pyi` file with this, it shadows the normal fabric code, and if i drop that inline in my code, then mypy complains that i'm overriding existing functions.

it seems therefore the only way to do this properly is to wrap the original `task` function so it would be really nice to see some movement on this PR...

i'll reroll with the tweaked signatures.

PR #2338, opened 2025-06-07
Visualization for the fabric codebase

ivanmilevtues:
In this PR, I have generated a high-level diagram for the fabric project. You can see how it looks when Github renders is here: https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/fabric/on_boarding.md

Me and a friend are trying to make it easier for maintainers and developers to get to know new codebases. And we think visualization is the way to go, that is why we are working on these diagrams. We think that the best way to start is via getting a high-level knowledge first and then dig deeper (into each component). We generate these diagrams by combining Static analysis and LLMs and this way allowing for multple generations which we hope will keep the whole document always up-to-date (probably via workflow automation?).

I would love to hear your opinion on the matter, any feedback is more than welcome!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/AppTaskProcessor.md
+++ b/.codeboarding/AppTaskProcessor.md
@@ -1,0 +1,48 @@
+```mermaid
+graph LR
+    Main_Program_Initializer["Main Program Initializer"]
+    Task_Execution_Engine["Task Execution Engine"]
+    Fabric_Exception_Handler["Fabric Exception Handler"]
+    Main_Program_Initializer -- "configures" --> Task_Execution_Engine
+    Task_Execution_Engine -- "depends on" --> Fabric_Exception_Handler
+```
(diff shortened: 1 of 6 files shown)


davidjmemmett:
I've had a very quick glance at what you've submitted & I'm not sure what you're trying to accomplish here. It looks like the typical rubbish we've all come to expect from LLM-based content generators; barely anything in these changes bares any semblance of reality.

From my perspective, these changes proposed don't serve any purpose other than to misinform and cause confusion.

ivanmilevtues:
Thank you for the answer @davidjmemmett!

The aim is to make a nice high-level diagram which represents the codebase, from your answer it seems like we are missing the point for your repository. Thank you so much for the input, I will take a close look into it and will try to adress the issues.

I wonder in general turms what do you think about a diagram driven documentation? Do you like the idea of having diagrams or you prefer better just a text document! Curious to catch your opinion on how you would go about on-boarding, as in the end of the day I don't want to just spam people, but build something that people are happy to merge and use!

davidjmemmett:
In my opinion, you're wasting your time here.

If any such diagrams were ever required, they'd be crafted by developers who are familiar with the codebase, guided by our own experiences & using freely-available open-source tooling.

ivanmilevtues:
Thanks for sharing, I think we will continue working on this as we already got a merge request accepted in an atlassian repository: https://github.com/atlassian-api/atlassian-python-api/pull/1543

This said I see your points and understand that it is not for everyone especially in the current state!

PR #2355, opened 2026-01-20
Correct behaviour of `ProxyJump none` in ssh_config

### LINKED ISSUE #2354: ssh_config `ProxyJump none` is not handled correctly
**Describe the bug**
Setting the ssh_config `ProxyJump` option to the value `none` should deactivate the feature (and prevent later entries in the config from applying). This is currently not the case, `none` is treated like a hostname, breaking the connection.

Ref: https://man.openbsd.org/OpenBSD-current/man5/ssh_config.5#ProxyJump

**To Reproduce**
Steps to reproduce the behaviour (please attach a minimal example):
1. Setup a `ssh_config` with a Host that includes `ProxyJump none`
2. ```python
   from fabric import Connection

   conn = Connection("host")
   if conn.gateway.host == "none":
       print("This should not happen")
   ```

**Expected behaviour**
Just like when using a normal `ssh` client, I expect `ProxyJump none` to disable the proxy jumping.

Data5tream:
This fixes #2354

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -500,6 +500,10 @@ class Connection(Context):
     def get_gateway(self):
         # SSH config wins over Invoke-style config
         if "proxyjump" in self.ssh_config:
+            # `none` is a special value that disables the proxy jump
+            # https://man.openbsd.org/OpenBSD-current/man5/ssh_config.5#ProxyJump
+            if self.ssh_config["proxyjump"] == "none":
+                return None
             # Reverse hop1,hop2,hop3 style ProxyJump directive so we start
             # with the final (itself non-gatewayed) hop and work up to
             # the front (actual, supplied as our own gateway) hop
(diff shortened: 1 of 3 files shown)


anarcat (approved):
LGTM, clean and simple. would just need an update to the changelog, but we might be able to live with that.

Data5tream:
@anarcat 
If this is going to be released directly, I can add an entry to the changelog. Otherwise I could add a „unreleased“ section (?)

BTW, the changelog on the website is outdated and does not include the latest release from the source file. Looks like the deployment was forgotten / didn’t went through.

anarcat:
> If this is going to be released directly, I can add an entry to the changelog. Otherwise I could add a „unreleased“ section (?)

yeah, i seem to recall there used to be such a section, but maybe jeff changed the way that works. I'd say this is good enough for now, thanks!

PR #2361, opened 2026-03-17
Fix #2264: Connection doesn't respect/load project config

### LINKED ISSUE #2264: Connection doesn't respect/load project config
**Describe the bug**
The Connection class ignores the `ssh_config_path` from project-level _fabric.yml_-file. The config is loaded to context correctly, but `Connection` does not pick it up.

**To Reproduce**
We have a fabric.yml at project/task level that contains `ssh_config_path: resources/config` and we have a task defined as:

```python
from fabric import Connection
from invoke import task

@task
def say_hello(ctx):
    with Connection("my-server.de") as cxn:
        cxn.run(f"echo Hello world!")
```

The ssh-config at `resources/config` looks like:
```
Host my-server.de
    User root
    IdentityFile ~/.ssh/operator.key
```

**Expected behaviour**
We expect fabric/Connection to logon as `root` with the given key file. Instead we get:
```
File "/mnt/c/Users/codebude/Documents/Git/fabric/.venv/lib/python3.11/site-packages/paramiko/transport.py", line 1640, in auth_publickey
    return self.auth_handler.wait_for_response(my_event)
```

If we pass the config from context explicitly to the Connection constructor the connection works fine:
```python
from fabric import Connection
from invoke import task

@task
def say_hello(ctx):
    with Connection("my-server.de", config=ctx.config) as cxn:
        cxn.run(f"echo Hello world!")
```

**Environment**
- Python 3.11.3
- Fabric 3.0.1
- Paramiko 3.1.0
- Invoke 2.1.1
- Client-OS: Windows 11
- Target OS: Ubuntu 20.04

danielalanbates:
Fixes #2264

## Summary
This PR addresses: Connection doesn't respect/load project config

## Changes
```
fabric/connection.py | 33 ++++++++++++++++++++++++++++++++-
 1 file changed, 32 insertions(+), 1 deletion(-)
```

## Testing
Please review the changes carefully. The fix was verified against the existing test suite.

---
*This PR was created with the assistance of Claude Haiku 4.5 by Anthropic | effort: high (extended thinking). Happy to make any adjustments!*

By submitting this pull request, I confirm that my contribution is made under the terms of the project's license (contributor license agreement).

### DIFF
--- a/fabric/connection.py
+++ b/fabric/connection.py
@@ -1,6 +1,7 @@
 from contextlib import contextmanager
 from io import StringIO
 from threading import Event
+import inspect
 import socket
 
 from decorator import decorator
@@ -44,6 +45,31 @@ def derive_shorthand(host_string):
         port = int(port)
 
     return {"user": user, "host": host, "port": port}
+
+
+def _get_invoke_context_config():
+    """
+    Try to retrieve the config from the current Invoke Context in the call stack.
+
+    This allows Connection objects to inherit configuration from the task's
+    context, including project-level fabric.yml settings.
+
+    Returns the config object if a Context is found in the call stack, or None
+    if no Context is found.
+    """
+    try:
+        for frame_info in inspect.stack():
+            frame_locals = frame_info.frame.f_locals
+            # Check for common context variable names used in tasks
+            for ctx_name in ("ctx", "context"):
+                if ctx_name in frame_locals:
+                    obj = frame_locals[ctx_name]
+                    if isinstance(obj, Context):
+                        return obj.config
+    except Exception:
+        # If anything goes wrong during inspection, just return None
+        pass
+    return None
 
 
 class Connection(Context):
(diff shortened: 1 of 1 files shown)


themavik (commented):
**What this PR does:** Fixes #2264 by having Connection inherit config from the current Invoke Context (e.g. fabric.yml) when config is not explicitly passed.

**Done well:** _get_invoke_context_config() walks the call stack for ctx/context and returns config when found—solves the project config inheritance gap. The fallback to Config() when no context exists preserves backward compatibility. The try/except around inspect.stack() avoids crashes if inspection fails.

**Concern:** Walking inspect.stack() on every Connection() call when config is None could be costly in hot paths. Consider caching per-call or documenting that this is intended for task-invocation contexts. Also, checking only ctx and context may miss custom variable names; worth noting in the docstring.

danielalanbates:
Thank you for your time reviewing this. We're withdrawing this PR. Apologies for any inconvenience.

PR #2368, opened 2026-06-23
Tolerate malformed or encrypted identity files when scanning SSH keys

HrachShah:
## Summary

`OpenSSHAuthStrategy.get_pubkeys` caught only `FileNotFoundError` around `PKey.from_path`, but `PKey.from_path` can raise several other classes for files that exist yet are not usable private keys:

- `ValueError` — bad PEM bytes or wrong password
- `OSError` — permission denied or non-file at the path
- `UnknownKeyType` — cryptography backend doesn't recognize the key
- `PasswordRequiredException` — encrypted key with no password supplied
- `SSHException` — other paramiko parse/load failures

A user with one such file in their identities list, ssh_config IdentityFile, or `~/.ssh/id_*` default locations would crash an entire connection attempt with an unhandled traceback instead of just skipping the bad file the way OpenSSH does.

## Fix

Introduce `_SKIP_KEY_ERRORS` covering all of these and emit a fabric debug log line naming the path and the underlying exception so users who care can still see why a key was dropped.

## Tests

Three new tests, one per source:
- `tests/auth.py::OpenSSHAuthStrategy_::get_pubkeys::fabric_config::silently_skips_malformed_and_encrypted_keys`
- `tests/auth.py::OpenSSHAuthStrategy_::get_pubkeys::ssh_config::silently_skips_malformed_and_encrypted_keys`
- `tests/auth.py::OpenSSHAuthStrategy_::get_pubkeys::implicit_user_home_locations::silently_skips_malformed_and_encrypted_keys`

Each exercises `ValueError`, `UnknownKeyType`, and `PasswordRequiredException` / `SSHException` in sequence and asserts the strategy still yields the one good key.

### DIFF
--- a/fabric/auth.py
+++ b/fabric/auth.py
@@ -9,8 +9,34 @@ from paramiko.auth_strategy import (
     InMemoryPrivateKey,
     OnDiskPrivateKey,
 )
+from paramiko.pkey import UnknownKeyType
+from paramiko.ssh_exception import (
+    PasswordRequiredException,
+    SSHException,
+)
 
-from .util import win32
+from .util import win32, debug
+
+# Errors we treat as "skip this identity file" when scanning identity paths.
+# PKey.from_path can raise:
+#   FileNotFoundError       — path doesn't exist
+#   OSError                 — permission denied, is-a-directory, etc.
+#   ValueError              — malformed PEM bytes / wrong password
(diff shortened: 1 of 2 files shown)


anarcat (approved):
LGTM, seems like a good idea.

PR #2369, opened 2026-06-23
reopen cached SFTPClient when its socket has been closed

### LINKED ISSUE #2367: Connection.sftp() can return a closed SFTPClient
If an sftp client is opened using `Connection.sftp()` and later closed, the fabric connection may return the closed sftp client on a subsequent call to `Connection.sftp()`.

**To Reproduce**
```python
conn = fabric.Connection(host="{}@{}".format('root', ''), connect_kwargs={"password": ''})
sftp_client = conn.sftp()
# Do stuff with sftp
sftp_client.close()

# Open a second client using the same connection
sftp_client = conn.sftp()

# OSError: Socket is closed exception raised here
sftp_client.open('my_file')
```

**Expected behaviour**
Connection should check if the SFTPClient it is about to return from `sftp()` is valid before returning it.

**Environment**
Should be environment agnostic.

**Additional context**
Relates to #1981 

**Suggested change**
```python
@opens
def sftp(self):
    """
    Return a `~paramiko.sftp_client.SFTPClient` object.

    If called more than one time, memoizes the first result; thus, any
    given `.Connection` instance will only ever have a single SFTP client,
    and state (such as that managed by
    `~paramiko.sftp_client.SFTPClient.chdir`) will be preserved.

    .. versionadded:: 2.0
    """
    # Recreate the client if it is closed
    if self._sftp is None or self._sftp.sock.closed:
        self._sftp = self.client.open_sftp()
    return self._sftp
```

HrachShah:
Fixes #2367.

``Connection.sftp()`` memoizes the `SFTPClient` it hands out, but it never checked whether the cached client's underlying socket was still alive. If user code called `.close()` on the client (or the network layer tore down the channel between two `Connection.sftp()` calls), every subsequent `sftp()` invocation returned the dead client and the first op against it raised `OSError: Socket is closed`.

The fix checks `self._sftp.sock.closed` alongside the existing `None` check, so the next call transparently opens a fresh client.

```python
@opens
def sftp(self):
    ...
    if self._sftp is None or self._sftp.sock.closed:
        self._sftp = self.client.open_sftp()
    return self._sftp
```

Tests:
- Updated the existing `lazily_caches_result` test to use `Mock()` sentinels with an explicit `sock.closed = False` so they're compatible with the new check.
- Added `reopens_when_cached_sftp_socket_closed`: caches one client, flips `sock.closed = True`, and asserts the next `sftp()` returns a fresh client and continues to cache it while the new socket stays open.

All 5 tests in the `sftp` group pass on Python 3.12. The 44 unrelated pre-existing failures (env issues with mocked SSH sockets, threading, stdin) reproduce on the unmodified upstream main.

### DIFF
--- a/fabric/auth.py
+++ b/fabric/auth.py
@@ -9,8 +9,34 @@ from paramiko.auth_strategy import (
     InMemoryPrivateKey,
     OnDiskPrivateKey,
 )
+from paramiko.pkey import UnknownKeyType
+from paramiko.ssh_exception import (
+    PasswordRequiredException,
+    SSHException,
+)
 
-from .util import win32
+from .util import win32, debug
+
+# Errors we treat as "skip this identity file" when scanning identity paths.
+# PKey.from_path can raise:
+#   FileNotFoundError       — path doesn't exist
+#   OSError                 — permission denied, is-a-directory, etc.
+#   ValueError              — malformed PEM bytes / wrong password
+#   UnknownKeyType          — backend doesn't recognize the key type
+#   PasswordRequiredException — encrypted key, no password supplied
+#   SSHException            — other paramiko parse/load failures
+# Together these cover every "this file isn't a usable private key" case, so
+# we silently skip rather than crash an entire connection attempt because one
+# of the configured identities (or one of the default ~/.ssh/id_* files) is
+# bad. Users who care can enable fabric debug logging to see the cause.
+_SKIP_KEY_ERRORS = (
+    FileNotFoundError,
+    OSError,
+    ValueError,
+    UnknownKeyType,
+    PasswordRequiredException,
+    SSHException,
+)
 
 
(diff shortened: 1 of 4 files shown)


fabric/auth.py
```diff
@@ -99,7 +125,8 @@ def get_pubkeys(self):
         for path in self.config.authentication.identities:
             try:
                 key = PKey.from_path(path)
-            except FileNotFoundError:
+            except _SKIP_KEY_ERRORS as exc:
+                debug(f"skipping identity {path!r}: {exc}")
```
anarcat, line 129: the general idea behind this patch is good, but i stopped reading here because this looks completely unrelated. please remove those unrelated hunks.

anarcat (changes requested):

